PR #23, opened 2014-12-30
set iOS 8 as minimum to allow App Store submission

pizthewiz:
App Store submission currently rejects binaries that include ObjectMapper because dynamic frameworks must have a deployment target of iOS 8 - ObjectMapper currently deploys to iOS 7.1.

When building a linker warning pops up that seems to hint at this:
```sh
ld: warning: embedded dylibs/frameworks only run on iOS 8 or later
``` 

And it turns out this is fatal at submission time, the App Store will reject binaries that include dynamic frameworks if those frameworks don't have a `minimumOSVersion` of iOS 8 or later. This isn't unique to ObjectMapper, other projects that have run into this and needed to support iOS 7 have elected to go to a static framework, see realm/realm-cocoa#900.


pizthewiz:
Any thoughts on this, is the project's intention to support iOS 7? In it's current configuration, ObjectMapper _cannot_ be used for apps submitted to the App Store. :cry:

tristanhimmelman:
Hey there, 
Thanks for bringing this up, I haven't actually run into this issue yet. I would love to support iOS 7 but I haven't had time to explore the best way to do this yet.

As a temporary solution, since the ObjectMapper source is quite small you could include the files directly (I know, not a fun solution :( ).

If you have suggestions, I'm open to suggestions!
Tristan

PR #32, opened 2015-01-21
Add struct support

ikesyo:
It requires adding `mutating` keyword to `map` func.


tristanhimmelman:
Nice work, thanks for the addition :+1:

ikesyo:
You, too :smiley:

PR #33, opened 2015-01-21
Support Carthage

eoinoconnell:
This adds support for Carthage: https://github.com/Carthage/Carthage

This is done by making your scheme shared.


tristanhimmelman:
If you have instructions for adding ObjectMapper using Carthage that I could add to the Readme file that would be appreciated :)

eoinoconnell:
Cool I'm on it

PR #43, opened 2015-01-27
fix typo and ad unit tests

corinnekrych:
This PR fixes Array/Dictionary of primitive types when doing mapping from json to object


tristanhimmelman:
Good find! thanks for the fix

PR #45, opened 2015-01-29
Refactor Mapper class

ikesyo:


ikesyo:
`mapArray(string JSONString: String)` now returns empty array when parsing failed.

ikesyo:
Style question: Which do you prefer tab or space for this project? @tristanhimmelman

tristanhimmelman:
@ikesyo I have been using tabs. Thanks for the contributions :+1:

ikesyo:
@tristanhimmelman Thank you, I got it. I'm going to change using tabs for this PR.

tristanhimmelman:
One thing I noticed is you updated two functions with code that looks like this (line 46 Mapper.swift):
```
let JSON = parseJSONDictionary(JSONString)
return JSON.map { self.map($0, toObject: object) }
```
It is very concise and I do like the use of the map function, but I think it the following might be easier to read:
```
if let JSON = parseJSONDictionary(JSONString){
    return map(JSON, toObject: object)
}
return nil
```
What do you think?

ikesyo:
Yep, the latter is more understandable and consistent with other codes. I've changed to that style.

Additionally Xcode project settings is changed for using tabs.

PR #52, opened 2015-02-04
Implement Transforms based on Protocol and type aliases

ikesyo:
This removes necessity of specifying Transform's type parameter.


tristanhimmelman:
Nice update :+1:

I was hoping to remove the Generic types from the transforms :smiley:

ikesyo:
My pleasure! :sparkles:

PR #54, opened 2015-02-05
Ikesyo refactor from json to json

tristanhimmelman:


ikesyo:
FYI: Xcode can parse triple-slash comments as documentation. :wink:

tristanhimmelman:
Hahaha oops :). Good to know

tristanhimmelman:
I am considering changing the map function definition in Mappable so that it doesn't conflict with the Swift map function. Do you have any suggestions on naming? I was thinking maybe one of the following:
```func mapping(map: Map)```
or
``` func mappable(map: Map)```

ikesyo:
I agree with it. 👍

I prefer the former of the two.

PR #58, opened 2015-02-09
Add map/mapArray/mapDictionary overload to Mapper that takes an AnyObject?

ikesyo:
Refactor FromJSON using it.


ikesyo:
Added also `mapDictionary` overload.

ikesyo:
Sorry, added `map` overload too.

PR #60, opened 2015-02-11
Swift 1.2

pizthewiz:
A couple of minor changes are required to build the framework and tests against Swift 1.2 and Xcode 6.3. I'm not sure what the Swift 1.2 plan is for ObjectMapper, but at the very least, a branch might be nice?

### DIFF
--- a/ObjectMapper/Core/Mapper.swift
+++ b/ObjectMapper/Core/Mapper.swift
@@ -265,7 +265,7 @@ public final class Mapper<N: Mappable> {
 			}
 
 			if let JSON = JSONData {
-				return NSString(data: JSON, encoding: NSUTF8StringEncoding)
+				return NSString(data: JSON, encoding: NSUTF8StringEncoding) as! String
 			}
 		}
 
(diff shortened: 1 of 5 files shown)


tristanhimmelman:
I hadn't made a plan for Swift 1.2 but a branch sounds like a good solution. I've merge your PR into a branch called swift-1.2.
Thanks for the fixes

tristanhimmelman:
This no longer seems necessary as of the latest release of iOS 8.2

pizthewiz:
Swift 1.2 is tied to Xcode 6.3/iOS 8.3 so I think this is still required as the master branch doesn't build with the Swift 1.2 compiler.

tristanhimmelman:
Ah I see, my mistake.

pizthewiz:
No worries, it was a bit odd to have two concurrent Xcode/iOS betas floating about.

pizthewiz:
I rebased to pick up the latest changes and added a couple of additional commits to address compile errors in the newer code.

In earlier commit comments, @ikesyo was thinking some tricky syntax could be used to avoid failable casts and in turn support both Swift 1.2 and earlier versions without supporting a separate branch - that should still be doable though now the code in question is more spread out.

pizthewiz:
Would it be helpful to rebase this again? I'm guessing with the release of Xcode 6.3 / Swift 1.2 today there will be more interest?

ikesyo:
By the changes in #89 and e271dc87b241176f958950158733c152d5e8417d, `master` branch is now compatible with Swift 1.2, I think.

pizthewiz:
It would appear so - closing this out.

PR #66, opened 2015-02-16
Can't map a generic array

hborders:
Demonstrates a bug where attempting to map to a generic array of a generic superclass whose type is specified explicitly by a subclass always yields nil.


tristanhimmelman:
Hi there, 
There we a couple issues with the code in your PR. I was able to get your test case working by doing the following:
- The Generic type within the WithGenericArray needed to conform to Mappable
- The JSON string passed into the testSubclassWithGenericArrayInSuperclass() function had a bug in the array format
I committed my changes and merged them into master if you want to take a look.
Tristan

hborders:
Thanks! I applied your suggested fix and my internal code works fine now.

PR #71, opened 2015-02-24
[WIP][POC] Split Mappable functionality to the immutable variant and the mutable one

### LINKED ISSUE #26: Non-optional mapping
I want to create an object from JSON, but if any of the properties aren't set successfully, e.g. the corresponding JSON value is missing or malformed, I want to return a nil object. The upside is that I can then assume that any object created from JSON has all of the required properties, and thus I don't need to use optionals. Is there a way to do this?

ikesyo:
It is a proof of concept, which possibly fixes #26. 

How do you guys think about the form :question:

### DIFF
--- a/ObjectMapper/Core/Mapper.swift
+++ b/ObjectMapper/Core/Mapper.swift
@@ -10,7 +10,7 @@ import Foundation
 
 public protocol Mappable {
 	mutating func mapping(map: Map)
-	init()
+	init?(_ map: Map)
 }
 
 public enum MappingType {
@@ -45,6 +45,14 @@ public final class Map {
 		currentValue = valueFor(key.componentsSeparatedByString("."), JSONDictionary)
 		
 		return self
+	}
+
+	public func value<T>() -> T? {
+		return currentValue as? T
+	}
+
+	public func valueOr<T>(defaultValue: T) -> T {
+		return (currentValue as? T) ?? defaultValue
 	}
 }
 
@@ -142,9 +150,10 @@ public final class Mapper<N: Mappable> {
 	/**
 	* Maps a JSON dictionary to an object that conforms to Mappable
 	*/
-	public func map(JSON: [String : AnyObject]) -> N {
-		let object = N()
-		return map(JSON, toObject: object)
+	public func map(JSON: [String : AnyObject]) -> N! {
+		let map = Map(mappingType: .fromJSON, JSONDictionary: JSON)
+		let object = N(map)
+		return object
 	}
 
 	//MARK: Mapping functions for Arrays and Dictionaries
--- a/ObjectMapper/Core/Mapper.swift
+++ b/ObjectMapper/Core/Mapper.swift
@@ -13,7 +13,7 @@ public protocol ImmutableMappable {
 }
 
 public protocol Mappable: ImmutableMappable {
-	mutating func mapping(map: Map) -> Self
+	mutating func mapping(map: Map)
 }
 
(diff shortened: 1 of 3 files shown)


tristanhimmelman:
This is an interesting approach. However I think it is a bit more complex then it needs to be. I want to make sure to keep the implementation of Mappable or (ImmutableMappable) as simple as possible. I will think about it some more though...

How about this strategy as an alternative:

We add a bool member variable to the Map object which monitors whether any mappings were nil or failed to be assigned to their respective field. Checking for nil could easily be done in the subscript function. To check if the values were assigned properly, the Map object would need to be passed into the FromJSON operators (which should be simple) and could simply check if the field was set at the end of each function. Alternatively the check could be done at the end of each Operator function.

Finally could add a new function to the Mapper class, or create a subclass which would handle immutable mappings by checking the state of the member variable in the Map object.

What do you think of this approach?

ikesyo:
I understand what you are describing, and it's much simpler way. But I think current `Mappable` definition can not handle the types which have some `let` properties:

```swift
struct Immutable {
	let prop1: String
	let prop2: Int
	let prop3: Bool
}
```

`Mappable`'s `init()` requirement blocks that such types will be correctly initialized. At least, it seems to that changing `init()` to `init?(map: Map)` (Failable Initializer with `Map`) or something like that (factory method in this PR) is required.

And also `<=` operator with `inout` parameter can not be used with `let` properties.

tristanhimmelman:
It seems this https://github.com/thoughtbot/Argo has a similar approach for handling immutable types.

I'm still debating whether this is something I would like to include. I like how simple the Mappable protocol is at the moment and I am hesitant to add support for immutable types if it means complicating the ObjectMapper interface.

ikesyo:
I changed the implementation to more simpler style. How about it? @tristanhimmelman 

Types which doesn't need immutable mapping can merely call `mapping(map)` in their `init?(_ map: Map)` impl. And also immutable types can support `mapping` function using temporary var, create a new value from them, then replace self with it (somehow like the test codes).

tristanhimmelman:
Nice, I like this approach :+1: 

Do you want to add some documentation on supporting Immutable types?

ikesyo:
@tristanhimmelman Thank you for accepting this :sparkles: 

Yes, I'd like to tackle it in another PR.

tristanhimmelman:
I just realized there may be an issue with the updates to Mappable protocol. Given a class structure as follows:
```
class Base: Mappable {
	var base: String?

	init() {}
	required init?(_ map: Map) {
		mapping(map)
	}
	func mapping(map: Map) {
		base <- map["base"]
	}
}

class Subclass: Base {
	var sub: String?

	override init() {
		super.init()
	}
	required init?(_ map: Map) {
		super.init(map)
		mapping(map)
	}
	override func mapping(map: Map) {
		super.mapping(map)
		
		sub <- map["sub"]
	}
}
```
If we call 
```
let object = Mapper<Subclass>().map(JSONObject)
```
the mapping function of the Base class will be called twice. Once when super.init? is called and once when super.mapping is called. Do you have any thoughts on how we can avoid this?

ikesyo:
@tristanhimmelman Thanks for pointing out it.

In this case, I think `Subclass`'s `mapping(map)` in `init?(_ map: Map)` is unnecessary. `mapping(map)` call in `Base`'s `init?(_ map: Map)` correctly calls subclass's implementation. I'll fix the test code soon.

PR #72, opened 2015-02-25
Updated operator from '<=' to '<|' which overrides the native less-than-equal-to

vpalivela:
Fix the issue with ambiguous operator error when linking. It was very confusing since the original was overriding the native "less than or equal to" operator.


tristanhimmelman:
Hey there, I have been thinking about changing the operator as well (so that it doesn't override the native operator). I didn't realize that is what was causing the "ambiguous operator" error though. I haven't fully decided on which operator I want to use though. I will either use your PR as is or alter it with a different operator. Thanks for the contribution :+1:

PR #75, opened 2015-03-07
Symetric mappings for nested keys

yuseinishiyama:
Nested keys like below work well for mapping JSON into an object

```swift
heightInCM <- map["height.value"]
```

It, however, doesn't work for mapping an object into JSON
```js
{
  "height.value" : 180,
}
```

An expected result is
```js
{
  "height" :  {
    "value" : 180
  }
}
```

### DIFF
--- a/ObjectMapper/Core/ToJSON.swift
+++ b/ObjectMapper/Core/ToJSON.swift
@@ -11,8 +11,20 @@ import Foundation
 class ToJSON {
 	
     func basicType<N>(field: N, key: String, inout dictionary: [String : AnyObject]) {
-		var temp = dictionary
 		var currentKey = key
+
+		// Nested keys
+		var allKeys = currentKey.componentsSeparatedByString(".")
+		if allKeys.count > 1 {
+			currentKey = allKeys.first!
+			allKeys.removeAtIndex(0)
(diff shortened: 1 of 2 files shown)


ObjectMapper/Core/ToJSON.swift
```diff
@@ -11,9 +11,21 @@ import Foundation
...
     func basicType<N>(field: N, key: String, inout dictionary: [String : AnyObject]) {
-		var temp = dictionary
 		var currentKey = key
 
+		// Nested keys
+		var allKeys = currentKey.componentsSeparatedByString(".")
+		if allKeys.count > 1 {
+			currentKey = allKeys.first!
+			allKeys.removeAtIndex(0)
+			let remainingKey = join(".", allKeys)
```
ikesyo: Why don't you using another array value like `join(".", Array(allKeys[1..<allKeys.count]))`, and you can remove `allKeys.removeAtIndex(0)` and use `let` for `allKeys` for more immutability.

ikesyo:
:+1: for this!

yuseinishiyama:
Thank you for pointing them out! :smile:
I've fixed them.

tristanhimmelman:
Nice addition!

PR #84, opened 2015-03-10
[ToJSON] Improve nested keys support

ikesyo:
In current implementation, multiple nested keys mappings with same level only preserve the last value.

This PR fixes the issue, and also refactors out nested keys mapping as a single function in `ToJSON`.


tristanhimmelman:
When I added support for nested keys in ToJSON I knew I wanted to refactor it in a more elegant way, but I didn't have the time. Thanks for this :smile:

ikesyo:
No problem! :grin:

PR #88, opened 2015-03-20
Create EnumTransform to transform into RawRepresentable Enums

kaandedeoglu:


kaandedeoglu:
This is an extension I've been using in my project to map from strings or ints into enum types in my model classes. I think it's pretty useful as a transform

tristanhimmelman:
Nice stuff! I've been wanting to add something like this

PR #89, opened 2015-03-21
Introduce Nimble for better assertion

ikesyo:
This enables avoiding force unwrapping or failable cast, and provides better assertion messages.

~~This PR uses a forked repository to use prebuilt framework binaries, without needs of preparing certificates in CI environment for building signed framework in it.~~


tristanhimmelman:
@ikesyo, Nimble looks like an cool library. 

I haven't used Carthage before, but from what I've read it can be used with or without submodules. I think it might be better if the Nimble framework was included (committed) directly to the ObjectMapper repository, that way it can be run immediately after checkout instead of requiring the developer to run `carthage checkout`. What do you think? 

I'm always a bit hesitant to include external frameworks in a project like this as I like to keep the complexity at a minimum. With that said, I think Nimble will be alright, as it only links to the Test target.

ikesyo:
@tristanhimmelman 

> I haven't used Carthage before, but from what I've read it can be used with or without submodules.

Yes, that is true. `carthage bootstrap` or `carthage bootstrap --use-submodules`.

>  I think it might be better if the Nimble framework was included (committed) directly to the ObjectMapper repository, that way it can be run immediately after checkout instead of requiring the developer to run carthage checkout. What do you think?

Is this about for the developers who would like to contribute the project? I prefer separating own dependencies handling to a manager like Carthage or CocoaPods, then use the frameworks built by them, not including the source codes directly.

But if you're not willing to do it, we can just use submodules managed by Carthage, and let's add the dependency's project to ObjectMapper's project or workspace as a subproject. In this case, dependencies are fetched automatically as submodules, and Xcode will automatically build dependent subprojects.

tristanhimmelman:
I don't mean including the Nimble source code directly.

When using Cocoapods, for example, if you commit the Pods directory to your repository you do not need to run "pod install" when pulling the project for the first time. Cocoapods still manages the dependencies, but the project is runnable for new developers.

Is it possible to commit just the frameworks that are built by Carthage so that new developers can run the project without running any extra scripts?

ikesyo:
Sorry, I misunderstood what you said :sweat_smile: 

Instead of committing built framework binary, I've set up a Xcode workspace which contains both ObjectMapper and Nimble project. Opening `ObjectMapper.xcworkspace`, developers can build the library and also can run tests (Nimble would be built implicitly by Xcode). Is it an acceptable way?

tristanhimmelman:
Thanks for the updates. 

I've updated the README doc with carthage instructions for getting the project ready to run tests. Please take a look to make sure I didn't miss anything.

ikesyo:
@tristanhimmelman Thank you for accepting it and the README update. I've set up the project with git submodule, so I think `carthage checkout --use-submodules` might be more suitable.

PR #115, opened 2015-04-29
Added support for NSString input

tiemevanveen:
NSString's are not automatically converted into Strings in swift 1.2. 
Therefore when you use the following code: 

    let myString = NSString(data: myData, encoding: NSUTF8StringEncoding)
    let myObject = Mapper<MyOject>().map(myString)

This method will be called and will return nil:

	/**
	* Maps a JSON object to a Mappable object if it is a JSON dictionary or NSString, or returns nil.
	*/
	public func map(JSON: AnyObject?) -> N? {
		if let JSON = JSON as? [String : AnyObject] {
			return map(JSON)
		}

		return nil
	}

While you would like to call: 

	/**
	* Map a JSON string to an object that conforms to Mappable
	*/
	public func map(JSONString: String) -> N? {
		if let JSON = parseJSONDictionary(JSONString) {
			return map(JSON)
		}
		return nil
	}

This took me 2 hours to figure out. 

It would be cool if a warning / compile error would be thrown to avoid a bug hunt like mine.
Another solution would be to support NSString, I've created a PR for this.

### DIFF
--- a/ObjectMapper/Core/Mapper.swift
+++ b/ObjectMapper/Core/Mapper.swift
@@ -163,11 +163,13 @@ public final class Mapper<N: Mappable> {
 	}
 
 	/**
-	* Maps a JSON object to a Mappable object if it is a JSON dictionary, or returns nil.
+	* Maps a JSON object to a Mappable object if it is a JSON dictionary or NSString, or returns nil.
 	*/
 	public func map(JSON: AnyObject?) -> N? {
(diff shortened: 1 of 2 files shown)


tristanhimmelman:
Hey there, sorry for the inconvenience. I had actually noticed this issue recently as well. Thanks for the fix.

tiemevanveen:
:+1: Thanks!

SpacePossum:
:+1:

PR #153, opened 2015-07-13
Just use SequenceType.flatMap(), remove Array.filterMap() implementation

ikesyo:
See http://swiftdoc.org/swift-2/protocol/SequenceType/#func-flatMap


ikesyo:
Thanks! :sparkles:

tristanhimmelman:
:+1:

PR #159, opened 2015-07-22
Accept Sets mapping and Xcode 7 beta 4 fix

Busta117:
hi, i see the need of accept Sets mapping for map core data entities, so... look and merge if it looks good

### DIFF
--- a/ObjectMapper/Core/Mapper.swift
+++ b/ObjectMapper/Core/Mapper.swift
@@ -8,7 +8,7 @@
 
 import Foundation
 
-public protocol Mappable {
+public protocol Mappable:Hashable {
 	init?(_ map: Map)
 	mutating func mapping(map: Map)
 }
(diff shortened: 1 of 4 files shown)


ObjectMapper/Core/Mapper.swift
```diff
@@ -8,7 +8,7 @@
 
 import Foundation
 
-public protocol Mappable {
+public protocol Mappable:Hashable {
```
ikesyo: I believe that `Mappable` must be `Hashable` is a bad decision. Instead, You can add methods below in generic extensions of `Mapper`:

```swift
extension Mapper where N: Hasable {
    public func mapSet(JSONString: String) -> Set<N> { ... }
    ...
}
```

ObjectMapper/Core/Operators.swift
```diff
@@ -345,4 +345,35 @@ public func <- <T: Mappable>(inout left: Array<T>!, right: Map) {
...
 	}
-}
\ No newline at end of file
+}
+
+
+// MARK:- Set of Mappable objects - Set<T: Mappable>
+
+/// Array of Mappable objects
+public func <- <T: Mappable>(inout left: Set<T>, right: Map) {
```
ikesyo: These functions should constrain the generic parameter as `<T: Mappable where T: Hashable>`.

ObjectMapper/Core/ToJSON.swift
```diff
@@ -121,7 +121,20 @@ internal final class ToJSON {
             objectArray(field, key: key, dictionary: &dictionary)
         }
     }
-    
+	
+	
+	class func objectSet<N: Mappable>(field: Set<N>, key: String, inout dictionary: [String : AnyObject]) {
```
ikesyo: These methods should constrain the generic parameter as `<N: Mappable where N: Hashable>`.

Busta117:
fixed with the changes suggested by @ikesyo

tristanhimmelman:
@Busta117 thanks for the addition :+1: 

It would be great if you could add a few tests as well to ensure everything keeps working well.

IgorBelyayev:
I'm unable to successfully deserialize a Set. Has anyone verified this behavior?

tristanhimmelman:
@IgorBelyayev it seems that sets of `Mappable` object function correctly, however, sets of primitive type are not currently working. I will try to add this functionality in the near future

Issue #156, opened 2015-07-20
Inability to use `let` / `struct`


zdavison:
Currently, because of the way ObjectMappers `<-` operator functions, there is no way to use `let` properties, and the `Mapper` class itself requires `AnyObject`, so where we'd often like to have immutable structs, we have mutable class instances.

Would be real nice if we could fix this limitation.

cezheng:
If you just don't want the properties to be public mutable, I guess you can make it private(set) like this
```
public private(set) var propertyName: TypeName
```

tristanhimmelman:
ObjectMapper does in fact support `let` type variables, however it requires some extra code to accomplish. There is an example which you can check out in ObjectMapperTests.swift:
https://github.com/Hearst-DD/ObjectMapper/blob/master/ObjectMapperTests/ObjectMapperTests.swift#L685

Otherwise there are other libraries that tackle this requirement more directly for example:
[Himotoki](https://github.com/ikesyo/Himotoki)

PR #161, opened 2015-07-27
Remove init? from mappable protocol

tristanhimmelman:


danhodos:
ObjectMapper is super useful; thanks for all your hard work!

If I'm reading this change correctly, it seems like we've lost the ability to define `let` variables _as well as_ the ability to specify non-optional (or implicitly-unwrapped optional) properties.

Seems like the former was useful for specifying proper immutability constraints (i.e. this data is read from the server, but code should not be allowed to change it) and the latter was useful for validating JSON (i.e. "this JSON is invalid", "this JSON is valid, but with incorrect keys", and "this JSON is valid, with correct keys, but the keys have the wrong types").

Both of those seem like useful and worthy things to me; I'm interested to know if the `UnsafeMutablePointer` technique that as being used was somehow bunk, or if there were other reasons for the change?

danhodos:
Oh! I just saw that others have been discussing the same question over in #163. Ha!

tristanhimmelman:
You can still specify implicitly unwrapped optionals.

Support for `let` variables was never official and required quite a bit of extra code to achieve. This is what is no longer supported based on my latest change. 

But as you see in #163 I am considering including a Map paramater in the `newInstance` func which would return the ability to validate based on the JSON.

ldiqual:
@tristanhimmelman 

> Support for let variables was never official and required quite a bit of extra code to achieve.

I'd be curious to know how to implement `let` support in my app, would you mind elaborating or linking to a code snippet?

tristanhimmelman:
@ldiqual please see this thread #156

ldiqual:
Hmmm, I see. Thanks @tristanhimmelman !

tristanhimmelman:
@ldiqual, no problem. The line number I referenced in that thread is actually no longer valid... See following link for `let` implementation:
https://github.com/Hearst-DD/ObjectMapper/blob/master/ObjectMapperTests/ObjectMapperTests.swift#L725

PR #186, opened 2015-08-31
Swift 2.0

benoit-pereira-da-silva:
In swift 2.0 with Xcode7 B6, static methods are **reputed finals!**
The "static func newInstance() -> Mappable" cannot be overridden anymore.

My proposition consists to pass the map to the initializer init?(_ map: Map). When a mappable object inheritates from another mappable object it should call super before proceeding to its own mapping. This approach is consistent with NSCoder pattern.

``` swift
// MARK: Mappable

required init?(_ map: Map) {
        super.init(map)
        mapping(map)
}
```

### DIFF
--- a/ObjectMapper/Core/Mapper.swift
+++ b/ObjectMapper/Core/Mapper.swift
@@ -73,7 +73,7 @@ private func valueFor(keyPathComponents: ArraySlice<String>, dictionary: [String
 			return nil
 
 		case let dict as [String : AnyObject] where keyPathComponents.count > 1:
-			let tail = dropFirst(keyPathComponents)
+			let tail = keyPathComponents.dropFirst()
 			return valueFor(tail, dictionary: dict)
 
 		default:
--- a/ObjectMapper/Core/Mapper.swift
+++ b/ObjectMapper/Core/Mapper.swift
@@ -9,7 +9,7 @@
 import Foundation
 
 public protocol Mappable {
-	static func newInstance() -> Mappable
+    init?(_ map: Map)
 	mutating func mapping(map: Map)
 }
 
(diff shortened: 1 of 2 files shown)


tristanhimmelman:
I am using Xcode 7 B6 as well but I am not seeing the issue you are pointing out. See https://github.com/Hearst-DD/ObjectMapper/tree/swift-2.0#subclasses for the syntax to you need use when subclassing Mappable classes.

I am, however, thinking of changing the protocol to pretty much what you are proposing. See branch protocol-update: https://github.com/Hearst-DD/ObjectMapper/tree/protocol-update.

benoit-pereira-da-silva:
@tristanhimmelman the new protocol you are proposing is matching mine and is perfect :+1: 

Concerning my previous issue, i have understood :  i was trying to override the "static func newInstance(map: Map) -> Mappable?" as declared in the protocol : 

```swift
public protocol Mappable {
	static func newInstance(map: Map) -> Mappable?
	mutating func mapping(map: Map)
}
```
I should have used  "override class func newInstance(map: Map) -> Mappable? "
Isn't it a little bit strange to declare a static func and override a class func ? M'i missing something ?

tristanhimmelman:
I agree it is a bit strange, but it is a work around. This is one of the reasons I want to change the protocol back to how it used to be.

tristanhimmelman:
I just merged my protocol-update branch into the swift-2.0 branch. 

Thanks for your input :+1:

benoit-pereira-da-silva:
I feel you are 100% right. And think we should adopt common patterns as possible.

``` swift 
public protocol NSCoding {
    
    public func encodeWithCoder(aCoder: NSCoder)
    public init?(coder aDecoder: NSCoder) // NS_DESIGNATED_INITIALIZER
}
```
Adopting the same approach will allow to convert seamlessly NSCoding compliant classes.

Issue #200, opened 2015-09-13
Memory leak when using nested objects


Torsph:
Hi!

Im having a problem with some memory leaks when using nested objects on iOS9 GM. 
Take the code under as an example:
```
let JSONString = "{\"username\":\"someusername\", \"age\": {\"age\":5}}"
let user = Mapper<User>().map(JSONString)

class Age: Mappable {
    var age: Int?

    required init?(_ map: Map){}

    // Mappable
    func mapping(map: Map) {
        age    <- map["age"]
    }
}

class User: Mappable {
    var username: String?
    var age: Age?

    required init?(_ map: Map){}

    // Mappable
    func mapping(map: Map) {
        username    <- map["username"]
        age         <- map["age"]
    }
}
```

If I run the code I will get memory leaks. If I remove the nested object the leak will go away.

It may because of this:
>
>Using switch against multiple types with as patterns may cause a memory leak. For example, avoid this kind of switch statement:
>```
>switch x {
>case let a as A: ...
> case let b as B: ...
> case let c as C: ...
> default: ...
>}
>```
>Rewrite the code to use if let a = x as? A statements instead of switch. This pattern performs type checks that avoid the memory leak. (22587077)
> ref: https://developer.apple.com/library/prerelease/watchos/releasenotes/DeveloperTools/RN-Xcode/Chapters/xc7_release_notes.html

That is used here: https://github.com/Hearst-DD/ObjectMapper/blob/master/ObjectMapper/Core/Mapper.swift#L94

tristanhimmelman:
Thanks, a fix has been merged into master #204

Torsph:
@tristanhimmelman thanks :) Could you possibly merge it into the swift2.0 branch? Or will you merge swift2.0 into master soon? :)

tristanhimmelman:
The Swift 2.0 branch has actually already been merged into master. You can use the latest release or point directly at master.

Torsph:
Ahh my bad. :)

PR #201, opened 2015-09-13
Fixes memory leak when using switch against multiple types with as pattern

Torsph:
Should close https://github.com/Hearst-DD/ObjectMapper/issues/200

Fixes memory leak when using switch against multiple types with as patterns that may cause a memory leak.

See https://developer.apple.com/library/prerelease/watchos/releasenotes/DeveloperTools/RN-Xcode/Chapters/xc7_release_notes.html under Swift for more info


tristanhimmelman:
Thanks for your PR. It has been merged via another PR which I created since I fixed a few other instances of the same problem :+1: 

https://github.com/Hearst-DD/ObjectMapper/pull/204

PR #216, opened 2015-09-21
Add description on mapping non-optional member.

yoshinorisano:
There is no description on how to map non-optional member, so I wrote about it.
If this is the right way to do that, please consider merge my pull request : )


yoshinorisano:
@tristanhimmelman Thanks for merge and polish my writing!

tristanhimmelman:
No problem! Thanks for the help :+1:

PR #259, opened 2015-10-23
Nested array mapping, required field parameter

tonyli508:
Support mapping toJSON for nested array mappings.

Nested array mapping is not strict mapping from JSON, still can use number like "0" as key. Like ["key" ["0': "abc"]].

For required field verification, we can define a global function, rewrite it in testing target into XCTAssert, what do you think?


tristanhimmelman:
I am closing this PR at the moment. It has never passed the continuous integration tests which is the first criteria I look at before accepting a PR.

Secondly, I don't really follow what you are adding here other than ToJSON for nested arrays

PR #261, opened 2015-10-26
Mappable, Array<Mappable, and Set<Mappable> initializers and toJSON

scottrhoyt:
Thanks again for the great framework! Because of the nature of the changes, I included a pretty lengthy discussion here. Since this mostly reflects my design ideas and not substantial functionality improvement, feel free to reject this PR if it doesn't fit your goals, but I thought making this available to you was the least I could do to say thanks. :) I don't offhand know why this can't be automatically merged if you accept though. I don't think I made any hard-to-merge changes.

# Mappable Extensions

## Purpose

This PR incorporates the `init` and `toJSON` additions to `Mappable`, `Array<Mappable>`, and `Set<Mappable>` that I implemented for a recent project. I believe this interface is a bit more _Swifty_ than the current interface and allows better encapsulation of the **ObjectMapper** framework.

## Changes

By using protocol extensions, I added failable initializers and a `toJSON()` for `Mappable`, `Array<Mappable>`, and `Set<Mappable>`. At a high level, these changes allow for the following use cases.

    class MyMappable : Mappable, Hashable { ... }

    let myMappable = MyMappable(JSONString: aJSONString)
    let myMappable = MyMappable(JSON: aJSONDictionary)

    let myArray = [MyMappable](JSONString: aJSONArrayString)
    let myArray = [MyMappable](JSONArray: aJSONDictionaryArray)

    let mySet = Set<MyMappable>(JSONString: aJSONArrayString)
    let myset = Set<MyMappable>(JSONArray: aJSONDictionaryArray)

    let myMappableJSON = myMappable?.toJSON()
    let myArrayJSON = myArray?.toJSON()
    let mySetJSON = mySet?.toJSON()

## Discussion

As I have been working on this project, I have found that ObjectMapper has made writing my JSON-based API client so much easier. However, the framework dependency had to be spread to a few places that I would prefer it didn't. This was mostly the API client classes as well as the model-testing classes. Since all my JSON-mapped model objects already had to include the framework to implement `Mappable`, this seemed like the ideal place to encapsulate all the framework knowledge. By removing these dependencies and localizing the knowledge to the model objects, changing mapping libraries or even serialization methodology in the future should be easier to accomplish. This functionality could be further abstracted into protocols (e.g. `JSONConvertible`) to make testability in the API client easier, though I didn't take it that far.

In addition, since Swift added failable initializers, the Apple frameworks have moved more towards this pattern instead of the factory pattern that has been prevalent. It seemed like a nice idea to implement this pattern for ObjectMapper as well for stylistic reasons. This was made quite a bit easier with protocol extensions.

## Caveats

I made the decision to not implement initializers that take `AnyObject?` arguments. While this might make sense for the calls to `Mapper` given the context, it seemed a bit misleading to provide such a vaguely typed initializer by default since there's a relatively small subset of the `AnyObject?` type that could successfully initialize an object. It also seemed to be fighting the type system a bit. I think a better solution is to leave that type check in the code that is initializing the `Mappable` where checking for having the right type to initialize is maybe more appropriate.

I was not able to extend this philosophy to `Dictionary<String, AnyObject>`. This was primarily due to the current limitation of not being able to conditionally extend protocols with non-protocol type requirements. There could be a way around this, but it escaped me.

I would also prefer to include the new initializers in the `Mappable` protocol as well so that implementers could chose to override the default functionality. However, I couldn't figure out how to do that without adding a bunch of `required` inits. Again, there could be a way around this. Also, if this functionality was refactored out into a `JSONConvertible` protocol, that would largely alleviate this concern. However, with current limitations of protocol extensions, you'd then run into the problem of not being able to do something like:

    extension Array : JSONConvertible where Element : JSONConvertible { ... }

## Testing

Tests are included to provide 100% coverage of the new extensions.

### DIFF
--- a/ObjectMapper.xcodeproj/project.pbxproj
+++ b/ObjectMapper.xcodeproj/project.pbxproj
@@ -8,6 +8,11 @@
 
 /* Begin PBXBuildFile section */
 		37AFD9B91AAD191C00AB59B5 /* CustomDateFormatTransform.swift in Sources */ = {isa = PBXBuildFile; fileRef = 37AFD9B81AAD191C00AB59B5 /* CustomDateFormatTransform.swift */; };
+		3BAD2C0C1BDDB10D00E6B203 /* Mappable.swift in Sources */ = {isa = PBXBuildFile; fileRef = 3BAD2C0B1BDDB10D00E6B203 /* Mappable.swift */; };
+		3BAD2C0D1BDDB10D00E6B203 /* Mappable.swift in Sources */ = {isa = PBXBuildFile; fileRef = 3BAD2C0B1BDDB10D00E6B203 /* Mappable.swift */; };
+		3BAD2C0E1BDDB10D00E6B203 /* Mappable.swift in Sources */ = {isa = PBXBuildFile; fileRef = 3BAD2C0B1BDDB10D00E6B203 /* Mappable.swift */; };
+		3BAD2C101BDDC0B000E6B203 /* MappableExtensionsTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 3BAD2C0F1BDDC0B000E6B203 /* MappableExtensionsTests.swift */; };
+		3BAD2C111BDDC0B000E6B203 /* MappableExtensionsTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 3BAD2C0F1BDDC0B000E6B203 /* MappableExtensionsTests.swift */; };
 		6A2AD0451B2C786C0097E150 /* Mapper.swift in Sources */ = {isa = PBXBuildFile; fileRef = 6AAC8FC419F048FE00E7A677 /* Mapper.swift */; };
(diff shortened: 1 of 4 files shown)


tristanhimmelman:
Hi @scottrhoyt, thanks a lot for the PR and the in depth discussion. Sorry about the slow response. 

I quite like your idea to extend the Mappable protocol to include init functions! I am currently in the midst of making some tweaks to your branch and will soon merge it into master.

Thanks a lot :+1:

scottrhoyt:
No problem! Let me know if you'd like any help.

scottrhoyt:
@tristanhimmelman I was wondering if you had given more thought to this change?

tristanhimmelman:
Hey @scottrhoyt, yes I have actually. I think I will merge it very soon. PR https://github.com/Hearst-DD/ObjectMapper/pull/328 includes your changes and is up to date with master. My only real concern is that unfortunately Xcode doesn't seem smart enough to provide code completion suggestions for the `Mappable` extension. If it weren't for that I would have merge this a while back....

scottrhoyt:
Awesome. Yeah, I've run into similar Xcode issues. Let me know if there is any way I can help!

PR #293, opened 2015-11-11
Optional Array and Dictionary types defaulting to nil instead of empty collection

ivanbruel:
I've been unit testing the fromJSON and toJSON of my models and I had an issue where an optional array would always have value `[]` instead of being `nil`.

```swift
struct Designer: Mappable {

  var identifier: Int?
  var name: String?
  var dresses: [Dress]?

  init(identifier: Int? = nil, name: String? = nil, dresses: [Dress]? = nil) {
    self.identifier = identifier
    self.name = name
    self.dresses = dresses
  }

  // MARK: JSON
  init?(_ map: Map) { }

    identifier <- map["id"]
    name <- map["name"]
    dresses <- map["dresses"]
  }

}
```

Even though dresses is `[Dress]?`, when the JSON has no `dresses` key it would default to `[]`, hence when converting back to JSON it would serialize to `dresses: []`, not generating the same JSON that was passed in the first place.

PSA: I might be missing some edge cases where `[]` and `[:]` actually make sense as default values.

### DIFF
--- a/ObjectMapper/Core/Operators.swift
+++ b/ObjectMapper/Core/Operators.swift
@@ -221,23 +221,23 @@ public func <- <T: TransformType>(inout left: [String: T.Object]!, right: (Map, 
 	}
 }
 
-private func fromJSONArrayWithTransform<T: TransformType>(input: AnyObject?, transform: T) -> [T.Object] {
+private func fromJSONArrayWithTransform<T: TransformType>(input: AnyObject?, transform: T) -> [T.Object]? {
 	if let values = input as? [AnyObject] {
 		return values.flatMap { value in
 			return transform.transformFromJSON(value)
 		}
 	} else {
-		return []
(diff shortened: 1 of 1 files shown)


tristanhimmelman:
Thanks for the PR. 

Based on the actual changes in the code, I'm guessing you were only experiencing this issue when the mapping in question was using a Custom transform?

ivanbruel:
Not really a Custom transform, the case here is any property which consists of an Optional Array will default to an empty Array instead of a nil value. This does not seem to be consistent with other optional values such as `Int?`, `String?`, etc. 
Only happens when mapping Optional Arrays with JSON which does not provide the value or comes nil.
Without this change any Optional Array will default to `[]` instead of `nil`.
This is especially important for cases where empty arrays have different meaning than just plain old empty arrays.

PR #307, opened 2015-11-27
assign currentValue on Map init

baydet:
Hi! Thanks for the great library!

I want be able to use mapping some kind of that
```
struct Object: Mappable  {
    var relationship: OtherObject?
    var field: String?
 ... 
    func mapping(map: Map) {
        relationship <- (map, ObjectTransform<OtherObject>())
        field <- map["field"]
    }
}
```

Could you look at my PR? Thanks!

### DIFF
--- a/ObjectMapper/Core/Map.swift
+++ b/ObjectMapper/Core/Map.swift
@@ -44,6 +44,7 @@ public final class Map {
 	public init(mappingType: MappingType, JSONDictionary: [String : AnyObject]) {
 		self.mappingType = mappingType
 		self.JSONDictionary = JSONDictionary
+		currentValue = self.JSONDictionary
 	}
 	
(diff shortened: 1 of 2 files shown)


tristanhimmelman:
I'm not sure I like the idea of assigning currentValue on init... perhaps we can just make JSONDictionary public, to allow you to use it as follows:
```
func mapping(map: Map) {
        relationship = Mapper<OtherObject>().map(map.JSONDictionary)
        field <- map["field"]
    }
```

baydet:
I like the idea of working with mapping only through `subscript`. I think with moving `JSONDictionary` to public ObjectMapper will lose integrity - I will be able to work with `Dictionary` without `Map` object at all.

Anyway, I agree that my PR not solve the problem completely. It needs deeper analysis.

Issue #309, opened 2015-11-27
No operator overload for `<-` where T is mappable


tapi:
This is useful in cases where the object relationship is not inline. 
This can happen for a number of reasons, we use it in cases where an object may show up in the response many times so we use a URI to reference that object and create a placeholder in our datastore. When we encounter the full payload we then update our placeholder.
Similar behaviour can be seen with links in [JSON API](http://jsonapi.org)

Id be happy to add this if you think its reasonable.

tristanhimmelman:
Hey @tapi, I follow the example you are giving, but I'm not exactly sure what functionality you want to add. If you look here https://github.com/Hearst-DD/ObjectMapper/blob/master/ObjectMapper/Core/Operators.swift#L259 
there is an overload for `<-` where T is `Mappable`

tapi:
Well crap, I completely missed the part that includes the transform.

tristanhimmelman:
Actually... it does exist:
https://github.com/Hearst-DD/ObjectMapper/blob/master/ObjectMapper/Core/Operators.swift#L120

tapi:
Except T does not conform to Mappable in that operator ;)

tristanhimmelman:
Ahh I see. In that case, feel free to add and make a pull request :+1:

tapi:
Interesting

`public func <- <T: Mappable, Transform: TransformType where Transform.Object == T>(inout left: T, right: (Map, Transform))`
and
`public func <- <T: Mappable, Transform: TransformType where Transform.Object == T>(inout left: T?, right: (Map, Transform))`
are both fine but

`public func <- <T: Mappable, Transform: TransformType where Transform.Object == T>(inout left: T!, right: (Map, Transform))`
segfaults the compiler

tristanhimmelman:
Hmmm that is odd...

I just copy pasted those functions into the Operators file and was able to build the project without a problem.

tapi:
What version of Xcode? Im running on 7.1.1

tristanhimmelman:
I'm on the same version.

I do however get a seg fault with the following code due to the FromJSON line:
```
public func <- <T: Mappable, Transform: TransformType where Transform.Object == T>(inout left: T!, right: (Map, Transform)) {
	let (map, transform) = right
	if map.mappingType == MappingType.FromJSON {
		let value: T? = transform.transformFromJSON(map.currentValue)
		FromJSON.basicType(&left, object: value)
	} else {
[...]

PR #316, opened 2015-12-04
Support mappable transforms

tapi:
This PR aims to address #309.

The implementation is done, just working out some test cases right now. I wanted to push and get a feeling for style, correctness, etc.
Each commit references a specific use case of `<-` so they can be looked at in isolation rather than as one big change set.


ObjectMapper/Core/Operators.swift
```diff
@@ -285,6 +285,44 @@ public func <- <T: Mappable>(inout left: T!, right: Map) {
 	}
 }
 
+// MARK:- Transforms of Mappable Objects - <T: Mappable>
+
+/// Object conforming to Mappable that have transforms
+public func <- <T: Mappable, Transform: TransformType where Transform.Object == T>(inout left: T, right: (Map, Transform)) {
+	let (map, transform) = right
```
tristanhimmelman:
```
public func <- <Transform: TransformType where Transform.Object: Mappable>(inout left: Transform.Object, right: (Map, Transform)) 
```
What do you think of this function definition instead? I recently updated the other transform operator function definitions to be more like this (using Transform.Object rather than T).

I don't have much more time to look over this PR today, but from first glance, it looks pretty solid.

On another note, I think I might have to soon split up the Operators file into a few files... its getting a little large.

tapi:
I saw the syntax change in the class clusters branch but wanted to wait so as to avoid conflicts.

I was also thinking about splitting up the files and possibly reducing each operator to a single function (without `inout`) to skip some of the boiler plate. Figured this was probably out of scope of this PR ;)

tristanhimmelman:
Sorry I never got back you, but so far this all looks good to me. I'll check back after you add tests.

Would you mind giving me an example use case of how you are planning to use this functionality

tapi:
I was thinking of something like JSONAPI, but that use case is turning out to be a bit contrived. 
Not sure if you recall but we don't inline our relationships in our JSON response, instead we have a URI reference for that relation ship. Later on in the payload we include all of the related objects identified by there URI.

An example payload would be something like 
```json
{
	"teams": [{
		"api_uri": "/dota2/teams/8",
		"full_name": "Team Secret",
		"short_name": "Secret",
		"country": "International",
	}, {
		"api_uri": "/dota2/teams/43",
		"full_name": "Mineski",
		"short_name": "Mski",
		"country": "Philippines",
	}],
	"games": [{
		"api_uri": "/dota2/games/2723",
		"game_number": 2,
		"game_time": "33:49",
		"minutes": 33,
		"seconds": 49,
		"status": "post-game",
		"roshan_respawn_timer": 269,
		"max_gold_advantage": 31245,
		"max_xp_advantage": 28281,
		"match_url": "/dota2/matches/1418",
		"dire_team_url": "/dota2/teams/8",
		"radiant_team_url": "/dota2/teams/43",
		"winning_team_url": "/dota2/teams/8"
	}],
	"matches": [{
		"api_uri": "/dota2/matches/1418",
		"competition_label": "The Frankfurt Major 2015",
		"start_date": "2015-11-17T10:00:00.000Z",
		"status": "post-match",
		"tie_match": false,
		"max_games": 3,
		"round_label": "Playoffs Upper Bracket",
		"team1_score": 2,
		"team2_score": 0,
		"competition_url": "/dota2/competitions/32",
		"team1_url": "/dota2/teams/8",
		"team2_url": "/dota2/teams/43",
		"winning_team_url": "/dota2/teams/8",
		"current_game_url": "/dota2/games/2723"
	}]
}
```

tapi:
JSONAPI turns out to be contrived since the relationships are still dictionaries which object mapper has no problem with. In both cases though the `MappableCluster.objectForMapping()` is needed so that stub objects can be fully hydrated when they are encountered later in the payload.

tristanhimmelman:
Gotcha, I see where this is going now. Thanks for the details. Off the top of my head it seems that most of this can be handled by `objectForMapping`, but I may be missing something.

tapi:
In our particular case we want "/dota2/teams/8" to map to a dehydrated team object even though its a string and not a dictionary. Does cluster handle this case? also how would it handle collections say a team has an array of player URIs?

tapi:
Initial res does basic JSON API Testing using the transform. Its a little contrived but it works for direct relationships and for arrays

tristanhimmelman:
Yeah I guess I hadn't considered the use case properly. I dont think `objectForMapping` would work for the url string as ObjectMapper does expect a dictionary for any Mappable object mappings

tapi:
Like I said it'd work great for actually mapping those relationships in but I feel like thats an exercise best left up to the reader.

tapi:
Would you prefer the actual use case instead of the JSONAPI case which is just illustrating that the transform can be used?

tristanhimmelman:
Yeah I think I prefer your use case. It is a bit more demonstrative of what you can do with this feature.

Just thinking out loud here... here's another approach you might be able to use:
``` 
func mapping(map: Map){
    let team1URL: String
    team1URL <- map["team1_url"]

    team1 = Team.loadTeamUsingURL(team1URL) // load stub model which is later loaded
}
```

tapi:
it would certainty work, I prefer the use of the transform though, its more expressive and the equivalent of `loadTeamUsingURL()` would need to be written for every model thats loaded in this way resulting in a lot of duplication. It would also require a lot of manual effort to process collections.

tristanhimmelman:
Yep those are good points. please proceed

tapi:
Alright, I changed up the use case and came up with examples for all the operator overloads, optional and otherwise. Let me know what you think.
The non optional and implicitly unwrapped test cases are kinda necessarily copy pasta so maybe keep an eye out for things I may have missed

tristanhimmelman:
Looks good to me. Thanks for the addition! :+1:

PR #318, opened 2015-12-06
Make code functional

ken0nek:
- use `reduce` not `for-loop`

### DIFF
--- a/ObjectMapper/Core/Mapper.swift
+++ b/ObjectMapper/Core/Mapper.swift
@@ -412,38 +412,31 @@ extension Mapper where N: Hashable {
 }
 
 extension Dictionary {
-	internal func map<K: Hashable, V>(@noescape f: Element -> (K, V)) -> [K : V] {
-		var mapped = [K : V]()
-
-		for element in self {
-			let newElement = f(element)
-			mapped[newElement.0] = newElement.1
-		}
-
-		return mapped
-	}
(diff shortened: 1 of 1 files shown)


ObjectMapper/Core/Mapper.swift
```diff
@@ -412,38 +412,31 @@ extension Mapper where N: Hashable {
 }
 
 extension Dictionary {
-	internal func map<K: Hashable, V>(@noescape f: Element -> (K, V)) -> [K : V] {
```
chainone: For such a low layer of abstraction, I don't think applying "reduce" everywhere is a good idea. Performance hurts.

chainone: Let me know if I am wrong.

tristanhimmelman: I have not measured the performance of reduce so I don't know if this is true. However I don't see a compelling reason to use it in this situation. I personally find the syntax is a bit clearer without it.

chainone: Take look at this blog:
http://airspeedvelocity.net/

"Your reminder that building arrays with reduce, while fun, is accidentally quadratic."

tristanhimmelman:
Thanks for the help but I've decided not to include this PR based on the above conversation.

PR #325, opened 2015-12-10
Array nested keys support

marceloqueiroz:
This adds support for mapping a specific array element.

Example:
    
     {
        "distances": [{
 		    "text": "102 ft",
 		    "value": 31
 	     }]
     }

Your map function will look like:
   
     func mapping(map: Map) {
        match <- map["distances.0"]
     }


tristanhimmelman:
Thanks for this. I pulled your branch and made a few modifications and then merged them into master. This should have been marked as closed automatically by github but it wasn't for some reason...

piresbruno:
thanks @tristanhimmelman

PR #328, opened 2015-12-13
Scottrhoyt master

tristanhimmelman:


scottrhoyt:
:+1:  So glad to see this make it in. Let me know if there is any other ways I can help, and keep up the great work! I'm currently about to use ObjectMapper with [SwiftYaml](https://github.com/behrang/YamlSwift) and some custom extensions to easily map Yaml to swift objects!

Issue #276, opened 2015-11-03
How to manage polymorphic object


elpic:
On the old version of ObjectMapper, we have a class func called newInstance. That method returns a Mappable object. 
I used to use that method to generate the corresponding instance of the type I want.

For example: 

```
{
  type: 'a'
}
```

This will generate an object of class A

```
{
  type: 'b'
}
```
This will generate an object of class B

A and B are the same API requests.

How can I do the same behavior with the new mapping protocol?

I will appreciate your response.
Thanks

tristanhimmelman:
Hi @elpic, when I changed the protocol to `init?` I didn't realize that people were using the `newInstance` function as you described above. I am considering reverting to the old protocol, however it is a significant change that will affect many people so I am taking my time with the decision.

You can achieve polymorphic mapping in the current design, however, it requires a bit more code. Given the following JSON:
```JSON
{
    "object1" : {
        "type": "dog",
        "name": "joey"
     },
    "object2": {
        "type": "cat",
        "name": "fluffy" 
    }
}
``` 
You can map the objects based on the type flag as follows:
```
class Response: Mappable {

    init?(_ map: Map){}

    func mapping(map: Map){
    
        var object1JSON: Dictionary<String: AnyObject>?
        object1JSON <- map["object1"]
        if let type = object1JSON?["type"] where type == "dog" {
             let dog = Mapper<Dog>().map(object1JSON)
        } 
    }
}
```
Hope this helps!

elpic:
Thanks

elpic:
@tristanhimmelman any news on this?

tristanhimmelman:
@elpic, unfortunately I haven't had the time to consider this any more. So for the time being I do not plan on changing this.

PR #331, opened 2015-12-15
Allow handling polymorphic object

elpic:
To be able to handle polymorphic objects the JSONDictionary should be
available. Otherwise the map function doesn't work.

### DIFF
--- a/ObjectMapper/Core/Map.swift
+++ b/ObjectMapper/Core/Map.swift
@@ -33,7 +33,7 @@ import Foundation
 public final class Map {
 	public let mappingType: MappingType
 	
-	var JSONDictionary: [String : AnyObject] = [:]
+	public var JSONDictionary: [String : AnyObject] = [:]
 	public var currentValue: AnyObject?
 	var currentKey: String?
 	var keyIsNested = false


tristanhimmelman:
I'm not sure I want to make the JSONDictionary public. You can already access the data as follows:
```
map["key"].currentValue
```
For an example of polymorphic classes see the https://github.com/Hearst-DD/ObjectMapper/blob/master/ObjectMapperTests/ClassClusterTests.swift

elpic:
Do you remember the staff you give me to make the polymorphic staff to work:

https://github.com/Hearst-DD/ObjectMapper/issues/276

If I can't access that value I can't make the mapping work. Please let me know if there is somthing wrong. Take a look at the example on the issue

elpic:
@tristanhimmelman please consider the following example:

```
[
  {type: 'a', property_a: 'a'},
  {type: 'b', property_b: 'b'}
]
```

tristanhimmelman:
If you look at the ClassClusterTests file that I linked above, you will see an example of how to handle this situation in a better way (when I answered your previous issue, the above functionality didn't exist). Look at the Vehicle class in that Test file. It implements MappableCluster and returns the correct subtype based on the data in map.

elpic:
Thank you

PR #332, opened 2015-12-15
Changed Mapper JSON parsing utils from private to static.

levieggert:


ikesyo:
Those are not changed to public, so you can't use the methods outside the module.

levieggert:
Just fixed.  Thanks ikesyo!

tristanhimmelman:
Thanks guys!

PR #369, opened 2016-01-14
Use correct Mapper type when using toJSONString in docs

AndreasHeiberg:


tristanhimmelman:
It is not necessary to pass in the type information when using the `toJSONString` function as the type gets inferred by the passed in object.

AndreasHeiberg:
Really I could swear that it didn't compile without it leading me to double check the docs, then make above change and then having it compile. I would have to double check some time and give you more information.

AndreasHeiberg:
Yeah I get "Ambiguous reference to member 'toJSONString'" until I add the type.

tristanhimmelman:
If you look through the tests for this project you will see many instances of toJSONString. No warnings or errors are raised. I've also tested it in another project and I am not see the issue. 

Are you using the latest version of xcode/swift?

AndreasHeiberg:
XCode 7.2 (7C68)
Swift version 2.1.1

nilswiersema:
Same issue here. @AndreasHeiberg, did you get it working?

AndreasHeiberg:
@nilswiersema well the solution is to do Mapper<Model> as I showed in the pull request.

As for why it isn't working as intended I have no clue.

nilswiersema:
For me it seems to be related to Optionals, even though I have already tried unwrapping it into a new variable with both 'if let' and the unwrap (!) operator. Still digging into this.

Mapper().toJSONString(SomeMappableClass()) seems to work fine though.

tristanhimmelman:
I believe your function will work if you change it to the following:
```
func foo<T: Mappable>(bar: T)
{
    let string = Mapper().toJSONString(bar)
}
```

nilswiersema:
I ended up using something along the lines of:

``` swift
func foo(bar: Mappable)
{
    let string = bar.toJSONString()
}
```

Easy answers... :smile:

Issue #375, opened 2016-01-25
currentValue == nil by default?


zulkis:
In order to support full power of <- operator we should set "currentValue = JSONDictionary" in the init method of Map.
So it will be possible to just map everything from the Map to the object which could support generics and other dynamic mapping based ob the Map subscription methods.

Something like that:
self.dataResponse <- map // current value == nil, but I am expecting to get JSONDictionary there. 

If there were some reasons why it should stay nil - could you clarify?

zulkis:
I want to achieve next behavior:
```
public protocol Fetchable : Mappable {
    func keyPathResponseValue() -> String // To map in the dataResponses
}

class FetcherResponse<T:Fetchable, Error> : Mappable {
    var error : Error?
    var dataResponse : T?
    
    internal required init?(_ map: Map) {
        
    }
    
    internal func mapping(map: Map) {
        if let _ = map["exception"].currentValue as? String {
            error = Error(map)
        } else {
            self.dataResponse <- map
        }
    }
}
```

So idea about that to just map the map to the object. But I don't have an access to the JSONDictionary and I cannot map it straightforward because currentValue is nil by default

zulkis:
I just realized that currentValue resets every time we trying to get valueForKey through subscription.
Why implementation was made this way not the other?

tristanhimmelman:
Closing this as I accepted your PR #379

PR #376, opened 2016-01-26
Update Map.swift JSONDictionary getter

zulkis:
Added possibility to get JSONDictionary as a root value

### DIFF
--- a/ObjectMapper/Core/Map.swift
+++ b/ObjectMapper/Core/Map.swift
@@ -72,6 +72,12 @@ public final class Map {
 		return self
 	}
 	
+	// MARK: Real JSON dictionary data
+    
+    	public func rootValue() -> [String : AnyObject] {
+        	return self.JSONDictionary
+    	}
+	
 	// MARK: Immutable Mapping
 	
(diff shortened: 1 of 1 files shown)


zulkis:
I tried to explain everything in https://github.com/Hearst-DD/ObjectMapper/issues/375

tristanhimmelman:
Hi @zulkis, I think I would prefer to just make JSONDictionary a public variable in this situation.

zulkis:
Hi @tristanhimmelman, it is a good solution, but still it is not preventing the situation with setting with dictionary from the outside(weird situation but still possible).
Actually I prefer to modify property but in that case of setting private setter/public getter - we have setValue function in ToJSON.swift that uses JSONDictionary as internal property...

From my side - it is not so important how to get it :) More important to obtain this possibility, because I don't really want to fork it(or in case of using AlamofireObjectMapper fork both)

Thank you!

ikesyo:
Just writing `public internal(set) var JSONDictionary` would be fine.

zulkis:
@ikesyo, you're absolutely correct!

Issue #383, opened 2016-02-08
Support for immutable variables 


tristanhimmelman:
I'm creating this issue as a place to discuss how we might support immutable variables.

ObjectMapper already has some rudimentary support of immutable types. There is some documentation hidden within the README which you can reference.

Sega-Zero:
#384 tries to solve this, but it breaks the ability to add `Mappable` conformance via extension. I solved the immutable mapping problem with a little bit hacky, but pretty much working code:
```swift
//somewhere else in code
struct SampleStruct {
    let avatar: String
    let name: String
}

//another swift file that gathers all models that needs to be Mappable
extension SampleStruct: ImmutableMappable {
    static func map(map: Map) -> SampleStruct? {
        var mapAvatar = ""; mapAvatar <- map["avatar"]
        var mapName = ""; mapName <- map["name"]

        return SampleStruct(avatar: mapAvatar, name: mapName)
    }
}

// MARK: private boilerplate code

protocol ImmutableMappable {
    static func map(map: Map) -> Self?
}

private struct MappableStub: Mappable {
    var map: Map?

    init?(_ map: Map) {
    }

    // Mappable
    mutating func mapping(map: Map) {
        self.map = map
    }
}

extension ImmutableMappable {
    static func mappedInstance(json: String) -> Self? {
        guard let stub: MappableStub? = Mapper().map(json), let map = stub?.map else { return nil }
        return self.map(map)
    }
}
```

Now all I need is call `let sample = SampleStruct.mappedInstance(json)`

mpdomas:
We use a more straightforward method.
First we introduce an extension for `Map` that adds `typedValue()` function:
```swift
import ObjectMapper

extension Mappable {
    mutating func mapping(map: Map) {}
}
extension Map {
    func typedValue<T>()          ->  T?  { var v: T?; v <- self; return v }
    func typedValue<T>()          -> [T]? { var v: T?; v <- self; return v }
    func typedValue<T:Mappable>() ->  T?  { var v: T?; v <- self; return v }
    func typedValue<T:Mappable>() -> [T]? { var v: T?; v <- self; return v }
}
```
Then our imutable variable can be initialized like so:
```swift
import ObjectMapper

struct Human: Mappable {
    init?(_ map: Map) {
        guard let
            _name: String = map["name"].typedValue(),
            _age:  Int =    map["age"] .typedValue()
            else { return nil }
        
        name = _name
        age  = _age
    }
    
[...]

PR #384, opened 2016-02-09
[Immutable] Make init throw + Transform & Mappable support

ldiqual:
@tristanhimmelman 

Related to https://github.com/Hearst-DD/ObjectMapper/issues/383

This PR adds various improvements over the current immutable mapping support (see commented section in https://raw.githubusercontent.com/Hearst-DD/ObjectMapper/master/README.md)

### Immutable mapping

Immutable mapping can now be written like this:

```
class Base: Mappable {
    let base: String
    let date: NSDate

    init(_ map: Map) throws {
        base = try map["base"].valueOrFail()
        date = try map["date].valueWithTransformOrFail(DateTransform())
        // And a lot more
    }

        // unchanged
    }
}
```

This required the following change in the `Mappable` protocol:

```
public protocol Mappable {
--	init?(_ map: Map)
++   init(_ map: Map) throws
```

This has a couple advantages:
 - There's no more need to call `map.isValid` at the end of the initializer.
 - You can exit the initializer early if one of the mappings failed.
 - You can add custom validation logic and throw an error if the mapping doesn't fit your needs.
 - The library can be extended to throw meaningful errors (like `couldn't map attribute "base" with type "Base", payload: "123"`). As of now it just throws a very basic `ErrorType` subclass (`MappingError`)

### Transforms & Mappable

Added support for the following immutable mapping types. `valueOrFail()` throws, `value()` returns an optional.
 - Basic type T, with `T, [T], [String: T]`
 - Transforms mapping via `valueWithTransform/valueWithTransformOrFail`. `T, [T], [String: T]`
 - Mappable mapping, with `T, [T], [String: T]`.

### ImmutableObjectTests

Moved the small `Immutable` test from `ObjectMapperTests` to `ImmutableObjectTests` and added all cases that were covered in this PR. We might want to break it down into multiple tests, but I went for simplicity for this first draft.

## Discussion

@tristanhimmelman This PR's main goal is to kick off a conversation about immutable mapping. While I think the interface is acceptable, it's more meant to be a starting point to discuss how the library could be modified to support such a feature.

I'd love your input on the following points:
 - What do you think about the `init` change? Do you think it's acceptable to break the interface for such a feature?
 - How about `mapping()`? Right now it's really awkward to have to switch on the mapping type, create variables and map them with an `inout` operator. I was thinking about the following:
  - Only allow `JSON => object` mapping via `init(map: Mapping)` and drop the `<-` operator
  - Add a `->` operator in `mapping(map: Mapping)` and rename that method to `serialize()`.

Also, if you'd rather go with another process to discuss immutability changes, just let me know! Thank you :)

### DIFF
--- a/ObjectMapper.xcodeproj/project.pbxproj
+++ b/ObjectMapper.xcodeproj/project.pbxproj
@@ -82,6 +82,9 @@
 		891804CE1C122AF000E5C3EE /* MappableTypesWithTransformsTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 891804CC1C122AF000E5C3EE /* MappableTypesWithTransformsTests.swift */; };
 		891804CF1C122AF000E5C3EE /* MappableTypesWithTransformsTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 891804CC1C122AF000E5C3EE /* MappableTypesWithTransformsTests.swift */; };
 		BC1E7F371ABC44C000F9B1CF /* EnumTransform.swift in Sources */ = {isa = PBXBuildFile; fileRef = BC1E7F361ABC44C000F9B1CF /* EnumTransform.swift */; };
+		C0873D5E1C698657008344EE /* ImmutableObjectTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = C0873D5D1C698657008344EE /* ImmutableObjectTests.swift */; };
+		C0873D5F1C698658008344EE /* ImmutableObjectTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = C0873D5D1C698657008344EE /* ImmutableObjectTests.swift */; };
+		C0873D601C698658008344EE /* ImmutableObjectTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = C0873D5D1C698657008344EE /* ImmutableObjectTests.swift */; };
 		CD16030A1AC023D6000CD69A /* ObjectMapper.framework in Frameworks */ = {isa = PBXBuildFile; fileRef = CD1602FF1AC023D5000CD69A /* ObjectMapper.framework */; };
 		CD1603181AC02437000CD69A /* ObjectMapper.h in Headers */ = {isa = PBXBuildFile; fileRef = 6AAC8F7B19F03C2900E7A677 /* ObjectMapper.h */; settings = {ATTRIBUTES = (Public, ); }; };
 		CD1603191AC02451000CD69A /* Mapper.swift in Sources */ = {isa = PBXBuildFile; fileRef = 6AAC8FC419F048FE00E7A677 /* Mapper.swift */; };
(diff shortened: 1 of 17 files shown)


tristanhimmelman:
Hi @ldiqual, thanks for your work and sorry for the slow response.

I think this is an interesting approach, specifically the change to `init`. I will spend some more time thinking about it as I have to be very careful when changing the protocol. But it seems from my initial observations that it wouldn't have too much impact on existing functionality.

With regards to the `mapping()` function and having a serializing operator only, I don't think I would move forward with this approach. I like the simplicity of having one place where all mapping information goes (unless you are using immutable types). Furthermore (and more importantly), this would remove the ability to map JSON to an existing object.

davidlawson:
@ldiqual your changes are working well for me. I also added a `valueOrFail` for sets:

	/// Set of Mappable objects
	public func valueOrFail<T: Mappable>() throws -> Set<T> {
		guard let set = Mapper<T>().mapSet(currentValue) else {
			throw MapperError.error
		}
		return set
	}

camina-apps:
+1

ldiqual:
@tristanhimmelman Any update on your review? I would love to start improving the PR by adding things like Set mapping and meaningful errors (eg: `Couldn't convert key "coordinates" to type String`), but I need to make sure you're ok with the interface change first.

Another point of discussion: I'm not sure it's technically possible, but how about renaming:
 - `valueOrFail()` to `value()`. There would be `value() -> T?` and `value() throws -> T`, which the compiler could choose according to the target type.
 - `valueWithTransformOrFail()` to `value(transform:)`. This includes the point above and tries to comply with the Swift API guidelines: https://github.com/apple/swift-evolution/blob/master/proposals/0006-apply-api-guidelines-to-the-standard-library.md

tristanhimmelman:
@ldiqual sorry for the slow response. I've been quite busy so I haven't had much time to consider making significant changes to this project. I do like your approach in the PR (and suggestions) but I haven't decided yet on whether to proceed. 

One of the main reasons I am hesitant is that I am considering another change to the `Mappable` protocol. Something like this:
```
public protocol Mappable {
	static func objectForMapping(map: Map) -> Mappable?
	mutating func mapping(map: Map)
}
```
Switching the `init?` function with `objectForMapping` would allow ObjectMapper to be used in an extension to a final class. It would also simplify integration with Realm. Both of these are issues that come up pretty frequently. 

If I took that approach, I think we could still incorporate your work by creating a protocol extension like the following:
```
public protocol ImmutableMappable: Mappable {
	init(_ map: Map) throws
}
```

myung:
Is just me who do not like throws?

pocket7878:
Any updates? I want to use valueWithTransform/valueWithTransformOrFail s!

scottdelly:
The `init throws` part here is really cool. I'm switching over from Decodable and having non-optional values and knowing when a JSON decoding error has occurred is mission critical.

I like Decodable's approach to the value() and valueOrFail() question. Basically when you declare a property as optional, you can assign to it using `try?` and that is all you need to do.

```Swift
class Base: Mappable {
    let base: String
    let date: NSDate
    let nextBase: String?

    init(_ map: Map) throws {
        base = try map["base"] //fails if missing
        date = try map["date"].valueWithTransform(DateTransform())// fails if missing
        nextBase = try? map["next_base"] // won't fail if missing
    }
}
```

Thanks guys!

tristanhimmelman:
@ldiqual I've finally had some free time to put thought into this and I'd like to move forward with the `init throws` approach that started here. I've created a new branch called `immutable-mappings` where I merged your code with the latest changes on master. If you are still up for it, I'd really appreciate your help with this (others are welcome to help out as well). 

I agree that for immutable variables, the current mapping operator is not that well suited. I'm considering creating a new operator that would be used just for the purposes of writing JSON. My only concern with this approach is that it will create a lot of duplicated code that may be tedious to maintain... 

Anyways, I think the protocol change is a good place to begin.

scottdelly:
@tristanhimmelman Any idea when this might get merged in? I'd really like to switch to ObjectMapper for decoding but my team and I are not comfortable doing that until ObjectMapper throws errors and does not return unsafe dummy values. Also, please see my response above for additional suggestion on making the code look a bit cleaner too :)

hebertialmeida:
Any updates? I'd really like to switch to ObjectMapper for decoding also.

ldiqual:
I've moved my main project to https://github.com/JohnSundell/Unbox to handle decoding. Encoding is done manually with SwiftyJSON. I don't really have anymore incentive to keep working on this PR so I'll have to pass. Feel free to use my code as you will!

tristanhimmelman:
The `immutable-mappings` branch is up to date and running well. It is based on the work of @ldiqual . Not all types are supported but most are.

If anybody would like to contribute, that would be great.

hebertialmeida:
@tristanhimmelman Is there a documentation for immutable structs?

tristanhimmelman:
@hebertialmeida unfortunately there is no documentation at the moment. However, you can check out `ImmutableObjectTests.swift` for working examples. Classes and Structs are likely handled the same way

hebertialmeida:
@tristanhimmelman Thanks, moved to Unbox (https://github.com/JohnSundell/Unbox) and Wrap (https://github.com/JohnSundell/Wrap).

devxoul:
I hope to use this feature in ObjectMapper :)

mapo80:
Hi,

I need to use Immutable Struct, I've seen there's a branch for this.
Can I use master version and try to use immutable struct?

Thanks.

devxoul:
@mapo80, I'm trying to make a new immutable implementation based on `master`. I think I'll make a PR in several hours.

devxoul:
New implementation: #592

devxoul:
Hey guys, I made another implementation for supporting immutables. Could you check for this(#592)? Thanks!

tristanhimmelman:
Closing this. I have a PR coming soon with support for immutable variables

PR #416, opened 2016-03-18
Fixed swift 3 warnings

Schaltfehler:
`++` / `--`, `var` Paramter and `typealias`in Protocols are marked as deprecated and will result in an compiler error in Swift 3.

[remove-pre-post-inc-decrement](https://github.com/apple/swift-evolution/blob/master/proposals/0004-remove-pre-post-inc-decrement.md)
[remove-var-parameters](https://github.com/apple/swift-evolution/blob/master/proposals/0003-remove-var-parameters.md)
[replace-typealias-associated](https://github.com/apple/swift-evolution/blob/master/proposals/0011-replace-typealias-associated.md)

### DIFF
--- a/ObjectMapper/Core/Map.swift
+++ b/ObjectMapper/Core/Map.swift
@@ -90,7 +90,7 @@ public final class Map {
 			return value
 		} else {
 			// Collects failed count
-			failedCount++
+			failedCount += 1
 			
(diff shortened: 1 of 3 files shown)


Schaltfehler:
Sorry, I will wait till Xcode7.3 is released. 
The Swift version on  Xcode 7.2 is using doesn't know `associatedtype`.

PR #426, opened 2016-04-01
add matchExactPath trigger to map subscript

dimala:
It forces to get value for provided keypath otherwise returns nil
For example:

If json

`
{
error = "Some error text"
}
`

and mapping for this json

`
errorText <- map["error.message"]
`

Actual result is "Some error text" which is wrong because requested keypath is different. 
if `matchExactPath == true` result will be nil

### DIFF
--- a/ObjectMapper/Core/Map.swift
+++ b/ObjectMapper/Core/Map.swift
@@ -60,18 +60,29 @@ public final class Map {
 	
 	public subscript(key: String, nested nested: Bool) -> Map {
 		// save key and value associated to it
+		return map(key, nested: nested, matchExactKeyPath: false)
+	}
+	
+	public subscript(key: String, matchExactKeyPath matchExactKeyPath: Bool) -> Map {
+		// save key and value associated to it
+		let nested = key.containsString(".")
+		return self.map(key, nested: nested, matchExactKeyPath: true)
+	}
+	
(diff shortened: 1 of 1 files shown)


tristanhimmelman:
Unless I'm missing something, what you describe above is what the `nested` value is used for in the subscript.

If you write `map["error.message", nested: true]`, the key that will be match will be `error.message` not `error` -> `message`

dimala:
I faced the problem when error message could be at "message" or "error.message" keypath. 
Also, don't ask me why :), error message could be at "error" keypath, which in my case shouldn't be mapped, but it will be because of "error.message". Mapper checks if value is a dictionary or an array and if it's not, it returns value at "error" keypath (in my situation it is String), but i expect nil from "error.message".

See the following code:

        var possibleMessage: String = ""
        possibleMessage <- map["message"]
        
        var possibleMessage2: String? = ""
        possibleMessage2 <- map["error.message"]
        
        message = [possibleMessage, possibleMessage2 ?? ""]
            .flatten()
            .filter { $0.length > 0 }
            .joinWithSeparator("\n")

if json will be `{ error = "some_text" }` then `message` will be equal "some_text", but as you can see in code above there is no `"error"` keypath defined in map. Thats why it would be good to have ability to force mapping to get value from full keypath but not from some part of it.

`map["error.message", nested: true]` do the same as `map["error.message"]`.

PR #429, opened 2016-04-06
Add mapping identifier to Map class.

shu-ua:
In my project i must deserialize object to JSON with different structure. Actually, i did not find any good decision of this problem. 

So, i added `mappingIdentificator` to `Map` class and added optional default parameter to all methods in `Mapper` class, which use protocol  `Mapping`  method `- mapping(map: Map)`.
In this method you can get mappingIdentificator and make your mapping more dynamical.

**For example** 
All methods can used with `mappingIdentificator` and without (in this case `mappingIdentificator = nil`).

`let user1 = Mapper<User>().map(JSONString)`
`let user2 = Mappper<User>().map(JSONString, mappingIdentificator: .SomeEnumObject)`

In protocol method you cane use `mappingIdentificator`.

```
func mapping(map: Map) {        
   switch(map.mappingIdentificator) {
        ...
     }
}
```

### DIFF
--- a/ObjectMapper/Core/FromJSON.swift
+++ b/ObjectMapper/Core/FromJSON.swift
@@ -48,8 +48,8 @@ internal final class FromJSON {
 	/// Mappable object
 	class func object<N: Mappable>(inout field: N, map: Map) {
 		if map.toObject {
-			Mapper().map(map.currentValue, toObject: field)
-		} else if let value: N = Mapper().map(map.currentValue) {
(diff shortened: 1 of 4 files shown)


tristanhimmelman:
Thanks for your work @shu-ua. I've decided not to merge this code as I have added a slightly more flexible solution. The Map object now has a MapContext property which. MapContext protocol which developers can implement to hold whatever information they may need.

PR #440, opened 2016-04-17
Add NSDecimalNumber transform

thii:
Not sure if this project still takes custom transforms but NSDecimalNumber is kind of mandatory if you're dealing with currency.


tristanhimmelman:
Thanks!

PR #474, opened 2016-05-24
Immutable with transforms & mappable

pocket7878:
Based on https://github.com/Hearst-DD/ObjectMapper/pull/384 works.
Built

- valueWithTransform
- valueWithTransformOr
- valueWithTransformOrFail
- value<T: Mappable>
- valueOr<T: Mappable>
- valueOrFail<T: Mappable>


pocket7878:
I found valueOrFail cause a BAD_ACCESS_ERROR when key not found.
And now I think better to throw error instead of hacking type system.

PR #541, opened 2016-08-10
Resort to a separate protocol for failable initializer

andrebraga:
It is effectively impossible to conveniently implement ObjectMapper for NSManagedObject subclasses with the initializer being part of the protocol. So I'm repeating the same trick of separating the part of the protocol with the initializer into a protocol of its own.

### DIFF
--- a/ObjectMapper/Core/FromJSON.swift
+++ b/ObjectMapper/Core/FromJSON.swift
@@ -46,7 +46,7 @@ internal final class FromJSON {
 	}
 	
 	/// Mappable object
-	class func object<N: Mappable>(inout field: N, map: Map) {
+	class func object<N: BaseMappable>(inout field: N, map: Map) {
 		if map.toObject {
 			Mapper(context: map.context).map(map.currentValue, toObject: field)
 		} else if let value: N = Mapper(context: map.context).map(map.currentValue) {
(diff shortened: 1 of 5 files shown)


tristanhimmelman:
Thanks for the contribution, but I don't want to modify the framework in this way for the sake of CoreData.

andrebraga:
Notice, the exactly same problem will happen for each and every class that must derive from another class and has "delicate" initialisation requirements. It is similarly difficult to adopt OM for Realm objects. It is effectively impossible to  adopt OM unless you have control over the implementation. The proposed changes allow adopting it via an extension.

tristanhimmelman:
I do agree that ideally ObjectMapper should be usable in just an extension. 

On second pass, I do like your approach to the problem (sorry, I think i reviewed this too quickly the first time). I'm going to put a bit more thought into it and get back to you.

andrebraga:
This is really appreciated. :smile:

tristanhimmelman:
@andrebraga I made a few updates and fixed the tests on this branch. All that has now been merged into master. I'm sure why Github did not close this automatically...

Thanks for your addition! 👍

andrebraga:
Yay, thanks! With the class cluster work my score for PR acceptance is looking great :smile: :smile:

Get the word out to Realm people, I think they'll love this!

PR #542, opened 2016-08-10
Use valueForKeyPath (KVC) when the key path does not contain integers.

jonmarimba:
This speeds-up the mapping of paths that are of the foo.bar.baz.bat variety as NSDictionary's valueForKeyPath is considerably faster than the (tail) recursive func otherwise in use. Since, presumably, most key paths are of this variety for most users, this should provide a nice speedup for most everyone.

(Tests pass)

### DIFF
--- a/ObjectMapper/Core/Map.swift
+++ b/ObjectMapper/Core/Map.swift
@@ -87,8 +87,20 @@ public final class Map {
 			isKeyPresent = isNSNull ? true : object != nil
 			currentValue = isNSNull ? nil : object
 		} else {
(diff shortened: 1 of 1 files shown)


jonmarimba:
@tristanhimmelman 

In our project, the opposite is most definitely true, especially mapping JSON on a device like an old iPod touch. We have a number of deep (foo.bar.baz) keypaths to map, though so perhaps that's the reason that we overcome the NSDictionary cast performance hit to gain an overall improvement.

PR #549, opened 2016-08-18
fixes for swift 3 xCode beta 6 update

EvilClay:


Urkman:
Would be great, if that could be merged soon...

joshpc:
Should some/most of the classes be `open` instead of `public`? Right now with everything as just `public` there's no flexibility at all. I have a separate fork that opens up the transforms, specifically, but I don't know what other people's use cases are like.

ObjectMapper/Core/Mapper.swift
```diff
@@ -334,18 +334,18 @@ extension Mapper {
 	public func toJSONString(_ object: N, prettyPrint: Bool = false) -> String? {
 		let JSONDict = toJSON(object)
 		
-        return Mapper.toJSONString(JSONDict, prettyPrint: prettyPrint)
+        return Mapper.toJSONString(JSONDict as Any, prettyPrint: prettyPrint)
```
joshpc: You can avoid all of these casts if we change the declaration: https://github.com/Hearst-DD/ObjectMapper/compare/swift-3...joshpc:swift-3?expand=1#diff-a9498ca83904e6eb2c05ea55fe3c5a1dR348

tristanhimmelman: That's not quite accurate. Some of the JSON objects are `[String: Any]` and others are `[[String: Any]]`

tristanhimmelman:
Thanks for the contribution 👍

PR #557, opened 2016-08-30
Compile time improvements

ivanbruel:
ObjectMapper has been taking a lot of time to compile so I added the `-Xfrontend -debug-time-function-bodies` to `OTHER_SWIFT_FLAGS` and discovered the problem.

The Swift compiler seems to be having issues with Type inference and the Nil Coalescing Operator, as such I fixed both and managed to save a lot of time as can be seen in the screenshots below:

Before fix:
screen shot 2016-08-30 at 16 45 25

After fix:
screen shot 2016-08-30 at 17 36 07

I'm sure a lot more stuff can be improved but these cases were the most time demanding ones.

For reference: [Regarding Swift build time optimizations](https://medium.com/@RobertGummesson/regarding-swift-build-time-optimizations-fc92cdd91e31#.9wc64v20a)


tristanhimmelman:
Thanks @ivanbruel, great improvement 👍

PR #560, opened 2016-08-31
Fix code about failable initializer in README

matouka-suzuki:
Hi,

map.JSONDictionary is [String : AnyObject].

So, I modified to use value()

Thanks


tristanhimmelman:
When checking the existence of a value, either method works. I am going to leave as is for now. Thanks for your input

Issue #497, opened 2016-06-22
URLTransform - handle URLs with spaces


rodrigo-lima:
I was testing a case where one of the JSON attributes is an URL string containing spaces.
For example:

```
{
    "contentId": 371923734, 
    "contentURL": "http://someserver.com/content/abcd/Vision Corporate Responsibility.html"
}
```

My mapping struct is something like this: 

```
struct ContentResource: Mappable {
    init?(_ map: ObjectMapper.Map) { }

    var contentId:Int
    var contentURL:NSURL?

    mutating func mapping(map: Map) {
        contentId  <- map["contentId"],URLTransform()
        contentURL <- (map["contentURL"],URLTransform())
    }
}
```

After debugging for a while, it turns out that the result of `contentURL <- map` response is `nil`, because `NSURL(string: "url with spaces")` returns `nil`.
We need to encode the URL string before calling *NSURL*. (see http://stackoverflow.com/a/24879794 )

The fix below seems to work for me, though there's some more discussion on the SO thread above on how to handle other chars ( + , & , etc.. ), for a more robust/complete fix

**URLTransform.swift**
```
public class URLTransform: TransformType {
    public typealias Object = NSURL
    public typealias JSON = String

    public init() {}

    public func transformFromJSON(value: AnyObject?) -> NSURL? {
        if let URLString = value as? String,
            // handle URLs with spaces (and other non-escaped chars)
            escapedURL = URLString.stringByAddingPercentEncodingWithAllowedCharacters(NSCharacterSet.URLQueryAllowedCharacterSet()) {
            return NSURL(string: escapedURL)
        }
        return nil
    }

    public func transformToJSON(value: NSURL?) -> String? {
        if let URL = value {
            return URL.absoluteString
        }
        return nil
    }
}
```

tristanhimmelman:
Thanks for the fix. I updated the transform

rodrigo-lima:
Hi thanks for looking into this!
My bad that I didn't update this with a slightly better version (coincidentally, I just found the bug this morning....)

Sometimes, the server JSON returns:
- `https://blah.com/somepath?fileName=abce space` -- this requires escaping
- `https://blah.com/somepath?someDate=2016-07-08T02%3A19%3A25.113-0700` -- this is good, already escaped on the server side

you might end up with double encoded %, if server returns something like
- `https://blah.com/somepath?fileName=abce space&someDate=2016-07-08T02%3A19%3A25.113-0700`
[...]

PR #566, opened 2016-09-06
Add a flag to disable encoding of strings in URLTransform

aydegee:
Following up on https://github.com/Hearst-DD/ObjectMapper/issues/497, the encoding of URLs introduced in 9b8163a isn't always desirable. For example, as mentioned in that issue, the string may already be encoded or may contain characters you don't want to be encoded. This is the case with URLs that contain a fragment portion where the `#` character should not be encoded.

This PR adds an optional flag to the`URLTransform` initializer so that the encoding step can be skipped before constructing the `NSURL`.

### DIFF
--- a/ObjectMapper/Transforms/URLTransform.swift
+++ b/ObjectMapper/Transforms/URLTransform.swift
@@ -31,15 +31,30 @@ import Foundation
 public class URLTransform: TransformType {
 	public typealias Object = NSURL
 	public typealias JSON = String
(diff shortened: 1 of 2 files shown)


tristanhimmelman:
Hey there, thanks for the addition 👍 . I merged this via command line and made a few changes. Not sure why Github didn't mark this as merged automatically...

aydegee:
No worries. thanks a lot

PR #589, opened 2016-09-22
making currentKey (read-only) public

heyzooi:


tristanhimmelman:
Thanks for your contribution but I have merge another PR that does the same thing more concisely
`public internal(set) var currentKey: String?`

PR #592, opened 2016-09-23
[Immutable] ImmutableMappable

devxoul:
Related to #383, #384

## Summary

* Supports Swift 3
* This PR doesn't break existing implementation.
* Adds `ImmutableMappable` to support immutables.
* `ImmutableMappable` has `init(map:) throws`.
* `ImmutableMappable` makes `func mapping(map:)` optional; basically supports transforming from JSON to Immutable object. Users needs to implement this function to support reverse transform (object to JSON).
    * `init(map:) throws`: JSON -> Object
    * `mapping(map:)`: Object -> JSON

## Example

**User**

```swift
struct User {
  let name: String
  let createdAt: Date
  let updatedAt: Date?
  let posts: [Post]

  // JSON -> Object
  init(map: Map) throws {
    name = try map.value("name") // throws an error if fails
    createdAt = try map.value("createdAt", using: DateTransform()) // throws an error if fails
    updatedAt = try? map.value("updatedAt", using: DateTransform()) // optional
    posts = (try? map.value("posts")) ?? [] // optional + default value
  }

  // Optional: Object -> JSON
    guard case .toJSON = map.mappingType else { return }
    
    var name = self.name
    var createdAt = self.createdAt
    var updatedAt = self.updatedAt
    var posts = self.posts

    name <- map["name"]
    createdAt <- (map["createdAt"], DateTransform())
    updatedAt <- (map["updatedAt"], DateTransform())
    posts <- map["posts"]
  }
}
```

**Usage**

```swift
let user = try User(JSONString: JSONString)
```

## Discussion

* Will it better to provide `toJSON(map:)` or something else instead of `mapping(map:)` for `ImmutableMappable`s?

### DIFF
--- a/ObjectMapper.xcodeproj/project.pbxproj
+++ b/ObjectMapper.xcodeproj/project.pbxproj
@@ -7,6 +7,13 @@
 	objects = {
 
 /* Begin PBXBuildFile section */
+		030114A91D95187600FBFD4F /* ImmutableMappble.swift in Sources */ = {isa = PBXBuildFile; fileRef = 030114A81D95187600FBFD4F /* ImmutableMappble.swift */; };
+		030114AC1D951A4F00FBFD4F /* ImmutableTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 030114AA1D95197100FBFD4F /* ImmutableTests.swift */; };
+		030114AD1D951A5300FBFD4F /* ImmutableTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 030114AA1D95197100FBFD4F /* ImmutableTests.swift */; };
+		030114AE1D951A5600FBFD4F /* ImmutableTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 030114AA1D95197100FBFD4F /* ImmutableTests.swift */; };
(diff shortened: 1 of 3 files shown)


devxoul:
Why test fails?

tristanhimmelman:
@devxoul sometimes I have to restart some of the tests in Travis for them to succeed... I have restarted the two that failed, I think they will probably resolve themselves.

From a quick glance your work looks good, thanks for contributing 👍. I will take some more time soon to look at it closely. Here are a couple things I'm considering:
1) Should we make the mapping function optional?
2) How could we improving the toJSON functionality for immutable variables
2) Do we need a separate protocol form ImmutableMappable? I do like that this doesn't break any current functionality, but I wonder if it would just be simpler to have `init(map: Map) throws` as the main init protocol.

devxoul:
@tristanhimmelman, thanks for your review! These are my answer:

1. `Mappable` should require `func mapping(map:)` because this is the only function which defines mapping relationship (`init()` is not designated for mapping). However, `ImmutableMappable` has an initializer for doing that. I thought it was OK to not implement the `func mapping(map:)` if you're not using toJSON functionality.
2. Well, we could imagine something like `map["key"] <- self.property` instead of `self.property <- map["key"]` to get rid of declaring local mutable variables. But neither can this remove the code repetition :-(
3. I really want to make the initializer of `Mappable` throwable. Why I made `ImmutableMappable` separately is because I thought you were worrying about the breaking API changes. How about using `ImmutableMappable` as an experimental feature? We can improve the detailed algorithms while giving users an opportunity to use immutables.

PR #594, opened 2016-09-25
Devxoul immutable

tristanhimmelman:


devxoul:
How about creating a new branch named `immutable-mappable` and merge #592 into that branch? This PR also could be re-targeted to that branch. The reason is to see diffs easily and make these PRs as small as possible.

tristanhimmelman:
@devxoul sounds good. I just created the branch

PR #598, opened 2016-09-28
New toJSON only operator (>>>)

tristanhimmelman:
Added a new operator for to JSON only mappings


tristanhimmelman:
@devxoul I added a new operator for toJSON only mappings. Do you have any suggestions on the operator definition `>>>`? (I'm still contemplating the choice...) Unfortunately `->` is not available. 

I think eventually I might renamed the original operator `<>` to display the multi directionality of mapping. I am hesitant to change it as I like to limit the code changes required for developers upgrading between versions.

If we did rename the original operator, then we would have something like:
`<>`: two way mapping
`>>>`: toJSON only mapping

devxoul:
@tristanhimmelman, could you please rebase `immutable-mappable` branch to make this PR small? Commits of #592 and #594 are included in this PR.

```bash
$ git pull --rebase origin immutable-mappable
$ git push origin immutable-new-operator --force
```

tristanhimmelman:
@devxoul I rebased for you.

Picking a custom operator is tough... I'll think about it some more. I think I will maintain the same order of parameters though.

PR #604, opened 2016-09-30
Add support for SPM

diogoguimaraes:
This pull request adds support for Swift Package Manager.

Thanks for your awesome work! 😄


tristanhimmelman:
Hi @diogoguimaraes, thanks for the PR. Unfortunately after merging the work on Immutable mappings, this branch no longer is mergeable. 

I would like to support swift package manager, so if you are willing to resolve the conflicts that would be great

diogoguimaraes:
Hey @tristanhimmelman, it's done 👍 
Can you please take a look and see if everything is ok?

Thanks

tristanhimmelman:
@diogoguimaraes thank you.

A few comments:
1) Can we maintain the current file structure within the `Sources` directory? I would prefer the files to be organized in folders that match the existing groups in Xcode.
2) It seems that your PR has modified the spacing within all of the source files. I would prefer if the PR makes necessary changes to the repository.

Sources/Mappable.swift
```diff
@@ -65,20 +65,20 @@ public extension BaseMappable {
...
+
 	/// Returns the JSON String for the object
 	public func toJSONString(prettyPrint: Bool = false) -> String? {
 		return Mapper().toJSONString(self, prettyPrint: prettyPrint)
 	}
 }
 
 public extension Array where Element: BaseMappable {
-	
+
```
tristanhimmelman: Example of the file modification that should not be included in the PR.

diogoguimaraes: I'll revert all these changes.

diogoguimaraes:
Regarding the file structure, AFAIK any folder inside the the `Sources` folder is considered a different module, so if you add folders the `swift build`will fail with the following error: `error: the package has an unsupported layout, unexpected source file(s) found`. So in order to support SPM I would leave the structure as it is.

diogoguimaraes:
@tristanhimmelman I messed up a bit in git and had to do a new fork to make a clean PR. I'll be closing this one and moving it to #618. 
Sorry about this mess 😞

PR #616, opened 2016-10-10
HEXColorTransform with testHEXColorTransform

vitkuzmenko:
Hello!
I create HEXColorTransform for all platforms with tests

### DIFF
--- a/ObjectMapper/Transforms/HEXColorTransform.swift
+++ b/ObjectMapper/Transforms/HEXColorTransform.swift
@@ -1,0 +1,99 @@
+//
+//  HEXColorTransform.swift
+//  ObjectMapper
+//
+//  Created by Vitaliy Kuzmenko on 10/10/16.
+//  Copyright © 2016 hearst. All rights reserved.
+//
+
+#if os(iOS) || os(tvOS) || os(watchOS)
+import UIKit
+#else
+import Cocoa
+#endif
+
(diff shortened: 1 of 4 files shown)


ObjectMapper/Transforms/HEXColorTransform.swift
```diff
@@ -0,0 +1,99 @@
...
+//  Copyright © 2016 hearst. All rights reserved.
+//
+
+#if os(iOS) || os(tvOS) || os(watchOS)
+import UIKit
+#else
+import Cocoa
+#endif
+
+open class HEXColorTransform: TransformType {
```
devxoul: I think it should be `HexColorTransform` rather than `HEXColorTransform`.

According to [Swift API Design Guideline](https://swift.org/documentation/api-design-guidelines/#follow-case-conventions), acronyms which doesn't commonly appear as all upper case should be treated as camel case.

> Acronyms and initialisms that commonly appear as all upper case in American English should be uniformly up- or down-cased according to case conventions. **Other acronyms should be treated as ordinary words**.

'Hex' doesn't commonly appear as all upper case because Wikipedia uses [Hex](https://en.wikipedia.org/wiki/Hex) as a title of document. How do you think of it?

devxoul:
By the way, I love this PR because this is much better than what I made in my project :+1:

tristanhimmelman:
Thanks! 👍

PR #617, opened 2016-10-10
Fix immutable mapping of JSON arrays

nmccann:
Hi,

I recently updated my project to take advantage of the new immutable support (thank you for adding it by the way), but found that JSON arrays (ex. `[[String: Any]]`) weren't mapping correctly.

It appears this is because execution ends up in `Mapper.map(JSON: [String: Any])` which only handles the `StaticMappable` or `Mappable` case - not `ImmutableMappable`.

My naive approach was to follow the example of `StaticMappable` and add support in that method for `ImmutableMappable`. However I saw that `ImmutableMappable` defines a number of extensions on `Mapper`, I suspect the correct approach would be to provide such an extension that supported arrays.

### DIFF
--- a/ObjectMapper/Core/Mapper.swift
+++ b/ObjectMapper/Core/Mapper.swift
@@ -170,7 +170,7 @@ public final class Mapper<N: BaseMappable> {
 		return nil
 	}
 
-	/// Maps a JSON dictionary of dictionaries to a dictionary of Mappable objects
+	/// Maps a JSON dictionary of dictionaries to a dictionary of Mappble objects
 	public func mapDictionary(JSON: [String: [String: Any]]) -> [String: N]? {
 		// map every value in dictionary to type N
 		let result = JSON.filterMap(map)
@@ -190,7 +190,7 @@ public final class Mapper<N: BaseMappable> {
 		return dictionary
 	}
 	
-    /// Maps a JSON dictionary of dictionaries to an existing dictionary of Mappable objects
+    /// Maps a JSON dictionary of dictionaries to an existing dictionary of Mappble objects
     public func mapDictionary(JSON: [String: [String: Any]], toDictionary dictionary: [String: N]) -> [String: N] {
 		var mutableDictionary = dictionary
         for (key, value) in JSON {
--- a/ObjectMapper/Core/Mapper.swift
+++ b/ObjectMapper/Core/Mapper.swift
@@ -102,6 +102,13 @@ public final class Mapper<N: BaseMappable> {
 			}
 		}
 
+		if let klass = N.self as? ImmutableMappable.Type {
+			if let maybeObject = try? klass.init(map: map) as? N, var object = maybeObject {
+				object.mapping(map: map)
+				return object
+			}
+		}
+
 		// fall back to using init? to create N
 		if let klass = N.self as? Mappable.Type {
 			if var object = klass.init(map: map) as? N {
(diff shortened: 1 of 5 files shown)


ObjectMapper/Core/Mapper.swift
```diff
@@ -102,6 +102,13 @@ public final class Mapper<N: BaseMappable> {
 			}
 		}
```
nmccann: For consistency I should add a comment like `// check if object is ImmutableMappable`, although I would argue that the code is self-commenting.

ObjectMapper/Core/Mapper.swift
```diff
@@ -102,6 +102,13 @@ public final class Mapper<N: BaseMappable> {
...
 
+		if let klass = N.self as? ImmutableMappable.Type {
+			if let maybeObject = try? klass.init(map: map) as? N, var object = maybeObject {
```
nmccann: I wanted to replicate the behaviour of the `StaticMappable` handling as closely as possible - if the initializer fails for some reason, execution will continue on to the "fallback" if statement below. Alternatively I could just return `nil`, which seems reasonable because I think `ImmutableMappable` and `Mappable` are mutually exclusive - you wouldn't have an object that conforms to both (and from my quick experiments, such a setup won't compile).

devxoul:
I intended to leave a comment on cd824806f47cd4181f27c47155d343ebcb956909 :(

ObjectMapper/Core/Mapper.swift
devxoul: How about creating methods in the same scope with https://github.com/Hearst-DD/ObjectMapper/blob/master/ObjectMapper/Core/ImmutableMappble.swift#L144-L158? You can easily add `Mapper` methods for `ImmutableMappable`s.

The reason I defined mapping methods as extensions(#592) is to separate `ImmutableMappable` implementations from existing code.

nmccann: I tried this out, and while it does make sense, it also seems like there will be a lot of duplication. In additional to the `Mapper` methods, the `Array` and `Set` convenience methods would have to be duplicated as well (if I am correct).

I also found that this alternative solution only works when the compile-time type of the object was `ImmutableMappable` - in my particular use case I have a manager class that took in request objects whose only generic constraint was `BaseMappable`. So it seems like I won't be able to interchange `Mappable` and `ImmutableMappable` classes easily. Not that I really needed to anyways, since all my mappable objects are immutable now.

devxoul: Hmmm. @nmccann @tristanhimmelman let's think about this:

1. `mapArray()` should throw error when there's invalid JSON model

    ```swift
    let array = [validJSON_1, invalidJSON_2, validJSON_3]
    let models = try mapArray(JSONArray: array) // should fail?
    ```

    With this approach, `mapArray()` can only handle `ImmutableMappable`s.

2. `mapArray()` should ignore invalid JSON model

    ```swift
    let array = [validJSON_1, invalidJSON_2, validJSON_3]
    let models = mapArray(JSONArray: array) // should return [model_1, model_3]?
    ```

    With this approach, `mapArray()` can handle any `BaseMappable`s.

tristanhimmelman: I think I prefer approach number one. I think that if you define an model as `ImmutableMappable` you expect a error to be thrown if something was not correctly map.

With that said, I think it makes sense to add `mapArray` to the `ImmutableMappable` `Mapper` extension because I think this should be a `throwing` method. Yes, it will cause some code duplication, but I think it will provide a better interface for developers.

nmccann:
Thank you for the quick turn around on this @tristanhimmelman!

PR #621, opened 2016-10-11
Make init of MapError public.

azeff:
I’m not sure if it was intentionally made internal.
I wanted to throw this error as well, but since `init` is internal and not in extension there is no way to construct `MapError` from client code.


devxoul:
I intended it for internal usage, but there's no problem if it become a public :)

PR #639, opened 2016-10-24
public dateFormatter in DateFormatterTransform

Busta117:
#640 
+ : making public dateFormatter in DateFormatterTransform to allow get it in subclasses


devxoul:
@Busta117, why do we need this? Could you provide us more detail use case?

Busta117:
in my particular case i have to check different formats of the ISO8601 standard, so, i created a subclass of a DateFormatterTransform and in the method `transformFromJSON` i check if it is possible convert the string into one of those ISO formats and then return the Date, so, in that subclass i need to access to the `dateFormatter` to create the Date

Issue #641, opened 2016-10-24
ISO8601DateTransform by regex


Busta117:
hi
will be great if the ISO8601DateTransform class validate all of the possibilities of a ISO8601 String, maybe by a regular expression

tristanhimmelman:
I commented on #644. The idea is good but in the PR the tests were failing. THanks

PR #644, opened 2016-10-24
ISO8601 dates matching by regex

Busta117:
issue #641 
- : validate every ISO8601 date format and map it into a valid date
- : method `testMoreISO8601FormatsDateTransform` to test different ISO8601 formats 

validate the string of the date by regex and create a valid format string and valid date string to create the Date with the DateFormatter


tristanhimmelman:
Sorry I cannot merge this because the tests are failing.

PR #648, opened 2016-10-26
Update URLTransform.swift 

etartakovsky:
Update URLTransform to accept allowed CharacterSet as an init parameter with previously used CharacterSet as default value for backward compatibility.

The reason for this change is that for URL's like this one: `http://dummyimage.com/200x200.png/5fa2dd/ffffff`

existing URLTransform produces the following:
`\"http:\/\/dummyimage.com\/200x200.png\/5fa2dd\/ffffff\"`

instead of: 
`"http%3A%2F%2Fdummyimage.com%2F200x200.png%2F5fa2dd%2Fffffff"`

To produce the latter result URL should be encoded with `CharacterSet.urlHostAllowed` instead of currently used `CharacterSet.urlQueryAllowed`. 

So i added `allowedCharacters` init param with `CharacterSet.urlQueryAllowed` default value to be able to change an allowed character set.

### DIFF
--- a/Sources/URLTransform.swift
+++ b/Sources/URLTransform.swift
@@ -32,6 +32,7 @@ open class URLTransform: TransformType {
 	public typealias Object = URL
 	public typealias JSON = String
 	private let shouldEncodeURLString: Bool
+    	private let allowedCharacters: CharacterSet
 
 	/**
 	Initializes the URLTransform with an option to encode URL strings before converting them to an NSURL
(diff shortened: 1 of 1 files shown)


devxoul:
@Flar49, could you please write a test for this?

Flar49:
@devxoul Sure, i'll do it tomorrow.

tristanhimmelman:
Closing this as it has been merged in another PR. Thanks for the contribution

PR #664, opened 2016-11-09
add doc Chinese support 

lacklock:
add doc Chinese support


ManWithBear:
Why you not translate full README?

lacklock:
i plan when ImmutableMappable release translte it.
And Installation i think every developers can understand it ,  doesn't need to translate.
If you think it's would be better translate full content, i would finish it all.

lacklock:
done!

tristanhimmelman:
thanks @lacklock 👍

PR #666, opened 2016-11-09
Fixes assertion failed on ImmutableMappable object

arctouch-marcelogobetti:
- The method Mapper.map(JSON:) -> N? fails an assertion when the
BaseMappable object implements the ImmutableMappable protocol, because
so far the only BaseMappable-derived protocols being considered were
StaticMappable and Mappable.
- This commit adds check for ImmutableMappable, preventing an assertion
failed in this case.

### DIFF
--- a/Sources/Mapper.swift
+++ b/Sources/Mapper.swift
@@ -104,6 +104,15 @@ public final class Mapper<N: BaseMappable> {
 				object.mapping(map: map)
 				return object
 			}
+		} else if let klass = N.self as? ImmutableMappable.Type { // Check if object is ImmutableMappable
+			do {
+				if var object = try klass.init(map: map) as? N {
+					object.mapping(map: map)
+					return object
+				}
+			} catch {
+				return nil
+			}
 		} else {
 			// Ensure BaseMappable is not implemented directly
 			assert(false, "BaseMappable should not be implemented directly. Please implement Mappable, StaticMappable or ImmutableMappable")


devxoul:
`ImmutableMappable` is designed to throw exceptions on mapping failure. So every immutable mappings should be used with `try` or `try?` keyword. I think it would be better to let users know that they should have used `try` keyword on immutable mappings.

arctouch-marcelogobetti:
@devxoul Probably I didn't understand your point, but from what I got: `ImmutableMappable`'s `mapping` comes from `BaseMappable`, where its signature is not marked with `throws`, so you'll get a warning if you add `try` on it. Anyway using `try?` on the `init` method would be a code improvement over what I've done here. I'll wait for your feedback to fix this further.

devxoul:
@arctouch-marcelogobetti, I thought you were using something like:

```swift
let user = Mapper<ImmutableUser>().map(JSON: jsonDictionary)
```

Could you please provide me some detail about what you expected? (or where you got that assertion message)

arctouch-marcelogobetti:
Hi @devxoul , the only line in my code where I make reference to `Mapper<T>` is this one:
`let movies = try Mapper<Movie>().mapArray(JSONArray: response.results)`
`Movie` is a struct that implements `ImmutableMappable`. But putting a breakpoint on this `map` method where the assertion failed, it seems that the problem is coming from this other line in my code:
`return self.provider.request(type).mapObject(GenericResponse.self)`
where I'm using `Moya/ObjectMapper` to map an `Observable<Response>` into an `Observable<GenericResponse>` (`GenericResponse` is another struct of mine that also implements `ImmutableMappable`).
This is `Moya/ObjectMapper`'s signature for mapObject<T>:
`public func mapObject<T : BaseMappable>(_ type: T.Type) -> RxSwift.Observable<T>`
So it is just requiring that the base protocol is implemented, thus ImmutableMappable is accepted, but then ObjectMapper internally recognizes only `StaticMappable` and `Mappable`, failing the assertion otherwise.

devxoul:
@arctouch-marcelogobetti, thank you for providing details. Moya-ObjectMapper is not currently  supporting ImmutableMappable mappings. https://github.com/ivanbruel/Moya-ObjectMapper/pull/28 is the patch for this issue and https://github.com/ivanbruel/Moya-ObjectMapper/issues/27 :smile:

arctouch-marcelogobetti:
That's great, @devxoul , thank you!
I'm closing this PR then. It still feels odd to me that the `map` method here has a message considering ImmutableMappable, but the code that comes right above it, does not. Why creating a new set of `mapObject` and `mapArray` functions specifically typed to `ImmutableMappable` is a better solution? I know it probably is, but wasn't able to find this information in the commit message so I'm asking here. Thanks once again!

devxoul:
@arctouch-marcelogobetti, it's because users don't have to change their existing code to support `ImmutableMappable`s.

arctouch-marcelogobetti:
Not sure if I understood. What kind of change do you mean? Applying your fix in Moya-ObjectMapper isn't a code change? What about other repositories that make use of ObjectMapper? Is Moya-ObjectMapper the only one that had a chance to access a method that fails an assertion whenever the class implementing `BaseMappable` does not implement either `Mappable` or `StaticMappable`? Thanks again!

devxoul:
@arctouch-marcelogobetti, I think I misunderstood your previous question:

> Why creating a new set of mapObject and mapArray functions specifically typed to ImmutableMappable is a better solution?

Did you mean why I created a separated function `mapObject` and `mapArray` rather than adding `if-else` statement?

arctouch-marcelogobetti:
No, I mean that you probably had 2 solutions for that problem:
1) considering `ImmutableMappable` inside the asserting method that doesn't consider this protocol
2) creating another set of `map<T>` method, specifically for `ImmutableMappable`, so that it (shouldn't) doesn't get into the generic `BaseMappable` existing one
You adopted number 2, which on a first look, seems a little more problematic to maintain than number 1. Number 1 would keep the `map<T>` methods generic and only internally handle the situation, which seemed to be previewed as per the assertion message, but wasn't really implemented (again, the message quotes Mappable, StaticMappable and ImmutableMappable, but the code does only account for the first two).

What happens if a 4th variant from `BaseMappable` is created? Will we create again a new set of methods for it, or create a new `else if` to consider it without failing the assertion?

Also, the assertion is stating that I should not implement `BaseMappable` directly, but this is misleading, because I didn't do that. I did implement `ImmutableMappable`.

devxoul:
@arctouch-marcelogobetti, thanks for the detailed explanation.

Totally agree with you. Of course we need an additional assertion message for `ImmutableMappable` (for whom used `map()` without `try` by mistake). I could approve this PR if it was just about adding assertion message for `ImmutableMappable`. But this PR contains the initialization of `ImmutableMappable` in non-throwing function: `map()`. This is wrong way to use of `ImmutableMappable`.

As you can see in [ImmutableMappable.swift#L153-L157](https://github.com/Hearst-DD/ObjectMapper/blob/master/Sources/ImmutableMappable.swift#L153-L157), every `map()` functions are declared separately from the existing `Mapper`. Because `map()` on `ImmutableMappable` should throw.

So I asked you the reason why you had faced wrong `map()` function, and it was because Moya-ObjectMapper didn't support `ImmutableMappable`. And this is why I chose the option 2.

Please feel free to ask about something you don't understand :smile::+1:

arctouch-marcelogobetti:
I get it now @devxoul , this is exactly the answer I was looking for. I felt it was something like that, so I immediately close my PR, I just wanted a confirmation. This is great. So I'll move my focus to fixing Moya-ObjectMapper and if you'd like, I can open a PR for ObjectMapper adding a new assertion and changing the message on the existing one. Thank you!

PR #667, opened 2016-11-10
Asserts that ImmutableMappable type doesn't call non-throwing map

arctouch-marcelogobetti:
Related to #666


tristanhimmelman:
Thanks

PR #670, opened 2016-11-13
Add RawRepresentable support for ImmutableMapper

lukewar:
Add additional `value(…)` function to handle mapping of RawRepresentable types.

### DIFF
--- a/Sources/ImmutableMappable.swift
+++ b/Sources/ImmutableMappable.swift
@@ -78,6 +78,11 @@ public extension Map {
 		}
 		return value
 	}
+	
+	/// Returns a RawRepresentable type or throws an error.
+	public func value<T: RawRepresentable>(_ key: String, nested: Bool? = nil, delimiter: String = ".", file: StaticString = #file, function: StaticString = #function, line: UInt = #line) throws -> T {
+		return try self.value(key, nested: nested, delimiter: delimiter, using: EnumTransform(), file: file, function: function, line: line)
+	}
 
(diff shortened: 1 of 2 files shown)


devxoul:
@lukewar, thanks for your work! How about using `EnumTransform` just like [this](https://github.com/Hearst-DD/ObjectMapper/blob/master/Sources/EnumOperators.swift)? We can just return `self.value(..., using: EnumTransform())`.

lukewar:
Hey, sure. Double checking if I understood you right. Do you propose to define operators for ImmutableMapper, that will throw?

devxoul:
@lukewar I mean, the function `func value<T: RawRepresentable>(key, nested, ...)` could be easily implemented if you used `EnumTransform`.

```swift
public func value<T: RawRepresentable>(_ key: String, nested: Bool? = nil, delimiter: String = ".", file: StaticString = #file, function: StaticString = #function, line: UInt = #line) throws -> T {
	return try self.value(key, nested: nested, delimiter: delimiter, using: EnumTransform(), file: file, function: function, line: line)
}
```

lukewar:
Oh, ok. I misunderstood you then :). Sure, I will update this.

lukewar:
Implementation is updated. TBH I don't know why travis is failing. All tests passed for those builds.

tristanhimmelman:
Thanks @lukewar 👍.

For some reason the travis builds fail regularly. All I had to do was restart them and it fixed the issue.

PR #691, opened 2016-12-06
emit Null into JSON if field is nil

traylewin:


tristanhimmelman:
Thanks @traylewin, I have made some changes and additions to your PR. See #697 

I will be merging that soon so I am closing this.

Issue #689, opened 2016-12-03
Map context forwarding


haritowa:
Hello, why don't you forward mapping context in [map](https://github.com/Hearst-DD/ObjectMapper/blob/master/Sources/ImmutableMappable.swift#L90) method, or provide way to customize it?

tristanhimmelman:
Hey @haritowa, good point. The MapContext should be forwarded for immutable mappings. 

If you have the time, a PR would be appreciated for this functionality.

Thanks!

haritowa:
Should it be forwarded only for Immutable mappings? I believe any mappable type can benefit from Context forwarding

tristanhimmelman:
I believe it is already forwarded for the other protocols

haritowa:
Ok, just double-check, pr will be available in a couple of hours

haritowa:
Create [PR](https://github.com/Hearst-DD/ObjectMapper/pull/695), but Travis check failed for strange reason. Maybe performance check failed, but i barely affect this.

PR #695, opened 2016-12-10
Forward Mapping Context for Immutable Mapping

haritowa:
Forward mapping context for nested properties/arrays/dictionaries during immutable mapping. Fix [issue](https://github.com/Hearst-DD/ObjectMapper/issues/689)

In addition - add context mapping tests both for BaseMappable & ImmutableMappable

### DIFF
--- a/Tests/ObjectMapperTests/MapContextTests.swift
+++ b/Tests/ObjectMapperTests/MapContextTests.swift
@@ -41,7 +41,9 @@ class MapContextTests: XCTestCase {
         super.tearDown()
     }
     
-    func testMappingWithContext() {
+	// MARK: - BaseMappable
+	// MARK: Single
+	func testMappingWithContext() {
 		let JSON = ["name": "Tristan"]
 		let context = Context(shouldMap: true)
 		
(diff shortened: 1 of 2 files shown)


Tests/ObjectMapperTests/MapContextTests.swift
```diff
@@ -70,6 +72,179 @@ class MapContextTests: XCTestCase {
...
+	// MARK: Single
+	func testImmatableMappingWithContext() {
+		let JSON = ["name": "Anton"]
+		let context = ImmutableContext(isDeveloper: true)
+		
+		let person = try? Mapper<ImmutablePerson>(context: context).map(JSON: JSON)
+		
+		XCTAssertNotNil(person)
+		
+		XCTAssertEqual(person?.isDeveloper ?? !context.isDeveloper, context.isDeveloper)
```
devxoul: Why did you use `?? !context.isDeveloper` here? If it was just for making `Bool?` to `Bool`, wouldn't it better using `?? false` or `person!.isDeveloper`? (we have ensured that the `person` is not nil from the test above)

devxoul: That makes sense. Cool!

tristanhimmelman:
Thanks @haritowa 👍 , looks good.

For some reason some tests fail occasionally on TravisCI. I just restarted them and everything was fine.

haritowa:
Nice, love this repo :)

PR #700, opened 2016-12-16
Master

benoit-pereira-da-silva:
Import Foundation to allow to Build ObjectMapper embedded in a macOS
CLI. If not Xcode 8.2 does not recognize NSNull and other base NS types.


wassimseif:
Doesn't build on mac using swift 3.0 or 3.0.1. Use of  unresolved identifier 'NSNull'

benoit-pereira-da-silva:
Wassim I  m not able to reproduce your problem. The fix works on any Targets on my side. It seems normal to import `Foundations`.

PR #752, opened 2017-02-16
Add UnsignedInteger/SignedInteger mapping operators

### LINKED ISSUE #751: Cannot mapper with UInt type
I created struct and make var number: Uint = 0
After I map with json it always return 0 as default

Here is json.
let info = ["id":"12345678", "number":1]

devxoul:
## Background

```swift
struct MyModel: Mappable {
  var value: UInt = 0

    self.value <- map["value"]
  }
}

let json: [String: Any] = ["value": 123]
MyModel(JSON: json)?.value // 0 (expected 123)
```

`<-` operator for mapping `UInt` uses [basic type mapping function](https://github.com/Hearst-DD/ObjectMapper/blob/master/Sources/Operators.swift#L44-L52). Basic type mapping uses [`Map.value()`](https://github.com/Hearst-DD/ObjectMapper/blob/2.2.3/Sources/Map.swift#L122-L124) function to get type-casted value], but `Int` cannot be casted to `UInt` using `as`.

## Workaround

There are two protocols that wrap integer types: `SignedInteger` and `UnsignedInteger`. This PR adds `<-` operator functions to support mapping of both protocols.

## Related Issues

* Thie PR resolves #751

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,5 @@
 language: objective-c
-osx_image: xcode8
+osx_image: xcode8.2
 
 env:
   global:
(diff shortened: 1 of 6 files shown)


devxoul:
Hmm test fails on Release build 🤔 I'll check it

devxoul:
I have no idea why tests are failing. Tests are succeeded in my machine 🤔🤔🤔

devxoul:
It fails on Xcode 8.0 environment (https://travis-ci.org/Hearst-DD/ObjectMapper/builds/202295374) but passes in Xcode 8.2 environment (https://travis-ci.org/Hearst-DD/ObjectMapper/builds/204318738) :(

tristanhimmelman:
@devxoul is this PR ready to go? or are you making updates?

devxoul:
@tristanhimmelman No problem! It's ready to go :)

PR #760, opened 2017-02-23
Use Xcode 8.2 in Travis CI

devxoul:
* Most of developers use latest SDK
* #752 fails on Xcode 8.0 but passes on Xcode 8.2


tristanhimmelman:
It seems that the Apple Watch test is failing. Can you look into that?

devxoul:
@tristanhimmelman Fixed it :tada: I'll update #752 when this PR is merged into master :)

PR #761, opened 2017-02-23
Allow for non-subscripted maps to be passed to TransformTypes

ethanjdiamond:
## Background

I've made a change to fix a bug where maps that are not subscripted do not have their values passed to TransformerTypes. 

There were two root causes of this issue:
1. A Map init function does not keep the passed JSON as currentValue, so it incorrectly states that it has none.
2. When the transform operator is run, the value of the map is only passed if it has a key present, regardless of if there is a value present.

Here is an example of when you would want this:

```swift
MyModel(JSONString: "[\"A\", \"B\", \"C\"]")

struct MyModel: Mappable {
  var value: String!

    self.value <- (map, LastItemTransformer())
  }
}
```

Say I made a transformer that transforms an array of values into whatever the last item is. Because I don't know the index of the last item, I can't use the . notation in the map subscript. However, I need to get the value from the whole map, and operate on it.

Without this fix, this would not work because the map wouldn't have a currentValue and the transformer is not passed the map's value anyway because the map has no present key.

### DIFF
--- a/Sources/Map.swift
+++ b/Sources/Map.swift
@@ -53,6 +53,7 @@ public final class Map {
 				
 		self.mappingType = mappingType
 		self.JSON = JSON
+		self.currentValue = JSON
 		self.toObject = toObject
 		self.context = context
 		self.shouldIncludeNilValues = shouldIncludeNilValues
--- a/Sources/TransformOperators.swift
+++ b/Sources/TransformOperators.swift
@@ -14,7 +14,7 @@ import Foundation
 public func <- <Transform: TransformType>(left: inout Transform.Object, right: (Map, Transform)) {
 	let (map, transform) = right
 	switch map.mappingType {
-	case .fromJSON where map.isKeyPresent:
+	case .fromJSON where map.isKeyPresent || map.currentValue != nil:
 		let value = transform.transformFromJSON(map.currentValue)
 		FromJSON.basicType(&left, object: value)
 	case .toJSON:
@@ -36,7 +36,7 @@ public func >>> <Transform: TransformType>(left: Transform.Object, right: (Map, 
 public func <- <Transform: TransformType>(left: inout Transform.Object?, right: (Map, Transform)) {
 	let (map, transform) = right
 	switch map.mappingType {
-	case .fromJSON where map.isKeyPresent:
+	case .fromJSON where map.isKeyPresent || map.currentValue != nil:
 		let value = transform.transformFromJSON(map.currentValue)
 		FromJSON.optionalBasicType(&left, object: value)
 	case .toJSON:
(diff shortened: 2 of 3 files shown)


tristanhimmelman:
Hmmm, I'm not sure whether this approach makes sense. My concern comes from the fact that we cannot rely on `map` without a subscript to always have the currentValue set as the passed JSON. For example, if there is another mapping before hand, I don't think your scenario will succeed because the currentValue will be set to the previous mapping...
```
struct MyModel: Mappable {
  var value: String!
  var value2: String!

    self.value <- map["key"]
    self.value2 <- (map, LastItemTransformer())
  }
}
```

I will think a bit more on this, but maybe something like `map.root()` would make more sense...

ethanjdiamond:
Thanks Tristan. Maybe there's another way to solve my problem. Basically what I'm trying to do is that I have a bunch of structs that follow a protocol that I'm trying to parse into an array. So something like this:


```swift
struct City: StaticMappable {
  var zoo: Zoo!

  static func objectForMapping(map: ObjectMapper.Map) -> BaseMappable? {
        return City()
    }

        noise <- map["zoo"]
    }
}

class AnimalTransformer: TransformType {
   func transformFromJSON(_ value: Any?) -> Animal? {
      if Cow.canParse(value) {
           return Cow.init(JSON: value)
      }

      if Lion.canParse(value) {
          return Lion.init(JSON: value)
     }
   }
}

struct Zoo: StaticMappable {
   var animals: [Animal]!

   static func objectForMapping(map: ObjectMapper.Map) -> BaseMappable? {
        return Zoo()
    }

        animals <- (map["animals"], AnimalTransfomer())
    }
}

protocol Animal: StaticMappable {
    static func canParse(value: Any?) -> Bool
}

struct Cow: Animal {
    var noise: String!

    static func canParse(value: Any?) {
       return value["type"] == "Cow"
   }

    static func objectForMapping(map: ObjectMapper.Map) -> BaseMappable? {
        return Cow()
    }

        noise <- map["noise"]
    }
}

struct Lion: Animal {
    var hasMane: Int!

    static func canParse(value: Any?) {
       return value["type"] == "Lion"
   }

    static func objectForMapping(map: ObjectMapper.Map) -> BaseMappable? {
        return Lion()
    }

        noise <- map["hasMane"]
    }
}

City.init(JSONString: "{city: {zoo: {animals: [{type: "Cow", noise: "Moo"}, {type: "Lion", hasMane: 1}]}}")
```

That didn't work without my bugfix though. To be fair though, I guess it is kind of sidestepping the problem that I'm reparsing the object every time. Maybe I could pass map directly into transformFromJSON instead of the value it's holding? I don't think this would be a problem if I was using classes, since I could just subclass animal, but I need to use structs and you can't subclass those.

PR #767, opened 2017-03-03
To help developers do something after Mapping has been completed.

croyneaus4u:
Have extended `BaseMappable Protocol` to include a function called `doAfterMappingCompleted`. This will help developers who want to perform operations after the Mapping has been completed.

### DIFF
--- a/Sources/ImmutableMappable.swift
+++ b/Sources/ImmutableMappable.swift
@@ -283,6 +283,7 @@ internal extension Mapper where N: BaseMappable {
 		if let klass = N.self as? ImmutableMappable.Type,
 			var object = try klass.init(map: map) as? N {
 			object.mapping(map: map)
+			object.doAfterMappingCompleted()
 			return object
 		}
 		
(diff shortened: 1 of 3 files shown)


devxoul:
Why not just implement on `mapping()`?

croyneaus4u:
Yes, we can implement stuff in the function `mapping(map: Map)` after we define the maps. But I believe the functions should be separated as per their usage. And as such I proposed this.

Please consider these:
- Sometime, we need to do more than just mapping of data after we have received response from API. We might need to implement some operations such as sorting or filtering of the data based on other parameters. So in my opinion, we should implement a way where user can do after mapping operations specifically rather than doing it in the `mapping(map: Map)` function.

- User may want to instantiate or save data to some other entity which is an instance of a class other than the one being mapped. So they can do it in the function that I have define.

More or less, this function will define a proper segregated space to perform whatever operations the user wants in a better readable and defined way.

tristanhimmelman:
Thanks for the contribution but I have decided not to include this as I want ObjectMapper to remain as simple as possible

PR #777, opened 2017-03-15
Increase NSDecimalNumberTransform precision

### LINKED ISSUE #776: Quantity 1.66 is not correct even using NSDecimalNumber
### Your JSON dictionary:

```json
{
  "quantity": 1.66
}
```

### Your model:

```swift
struct Repo: Mappable {
  var quantity: NSDecimalNumber?

  init(_ map: Map) {
    quantity <- (map["quantity"], NSDecimalNumberTransform())
  }
}
```

### What you did:

```swift
let repo = Mapper<Repo>().map(myJSONDictionary)
```

### What you expected:

I exepected something like:

```swift
Repo(quantity: 1.66)
```

### What you got:

```swift
Repo(quantity: 1.6599999999999995904)  // expected the quantity is 1.66
```

Using NSDecimalNumber is no better than using double.

Jake00:
NSDecimalNumber currently eagerly transforms to floating point when presented with a JSON number which loses precision. This pull request solves that by using Foundation to do the conversion for us by attempting to cast the number to NSNumber first and grabbing the decimal value out of it.

I added tests which failed before the change and passed after, specifically: 
```
XCTAssertEqual(mappedObject?.decimal, NSDecimalNumber(decimal: decimal))
XCTAssertEqual failed: ("Optional(1.6599999999999995904)") is not equal to ("Optional(1.66)")

XCTAssertEqual(mappedObject?.decimal?.stringValue, decimalString)
XCTAssertEqual failed: ("Optional("1.6599999999999995904")") is not equal to ("Optional("1.66")")
```

This should also resolve #776.

### DIFF
--- a/Sources/NSDecimalNumberTransform.swift
+++ b/Sources/NSDecimalNumberTransform.swift
@@ -37,8 +37,9 @@ open class NSDecimalNumberTransform: TransformType {
     open func transformFromJSON(_ value: Any?) -> NSDecimalNumber? {
         if let string = value as? String {
             return NSDecimalNumber(string: string)
-        }
-        if let double = value as? Double {
+		} else if let number = value as? NSNumber {
+			return NSDecimalNumber(decimal: number.decimalValue)
+		} else if let double = value as? Double {
             return NSDecimalNumber(floatLiteral: double)
         }
         return nil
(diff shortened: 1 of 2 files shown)


tristanhimmelman:
Thanks!

joeymok:
Hi @Jake00 , Please take a look at issue #799. Any idea? Thanks.

PR #780, opened 2017-03-21
Add support to ImmutableMappable can become a property of Mappable

### LINKED ISSUE #668: ImmutableMappable as a property in Mappable
I've been trying to have a clear example where Mappable and ImmutableMappable co-exist in a project to make sure my extensions for `Moya` work properly.

I have a [Repository](https://github.com/ivanbruel/Moya-ObjectMapper/blob/master/Sample/Demo/Repository.swift) which implements `Mappable` and an [Owner](https://github.com/ivanbruel/Moya-ObjectMapper/blob/master/Sample/Demo/Owner.swift) which implements `ImmutableMappable`.

As far as I can tell these two should be able to co-exist in the same serialization hierarchy.

I seem to be hitting a wall when it comes to the type checking in the `Mapper` class.

```
assertion failed: 'ImmutableMappable' type requires throwing version of function map(JSON:)  - use 'try' before map(JSON:): file /Users/ivanbruel/Documents/Moya-ObjectMapper/Sample/Pods/ObjectMapper/Sources/Mapper.swift, line 108
```

What's the reasoning behind not letting `map(JSON:)` work with `ImmutableMappable` properties?

For sake of issue completion, by changing the function into the code below I can make it work.

```swift
/// Maps a JSON dictionary to an object that conforms to Mappable
public func map(JSON: [String: Any]) -> N? {
	let map = Map(mappingType: .fromJSON, JSON: JSON, context: context)
		
	if let klass = N.self as? StaticMappable.Type { // Check if object is StaticMappable
		if var object = klass.objectForMapping(map: map) as? N {
			object.mapping(map: map)
			return object
		}
	} else if let klass = N.self as? Mappable.Type { // Check if object is Mappable
		if var object = klass.init(map: map) as? N {
			object.mapping(map: map)
			return object
		}
	} else if let klass = N.self as? ImmutableMappable.Type { // Check if object is Mappable
		if let object = try? klass.init(map: map) as? N {
			return object
		}
	} else {
		// Ensure BaseMappable is not implemented directly
		assert(false, "BaseMappable should not be implemented directly. Please implement Mappable, StaticMappable or ImmutableMappable")
	}
	
	return nil
}
```

devxoul:
## Background

`ImmutableMappable` was originally designed to not be used as property of the `Mappable`. It's because the `map()` function on `BaseMappable` cannot throw an error if `ImmutableMappable` is not correctly mapped. But this behavior can occur many confusions because there's no compiler level warnings.

I think `NSException` can raise an runtime exception if there's an error on mapping. It's important to make application safe, exceptions are only raised in debug environment.

Any discussions are welcomed 😃

## Related Issues

* This PR resolves #668

### DIFF
--- a/Sources/Mapper.swift
+++ b/Sources/Mapper.swift
@@ -106,8 +106,22 @@ public final class Mapper<N: BaseMappable> {
 				object.mapping(map: map)
 				return object
 			}
-		} else if N.self is ImmutableMappable.Type { // Check if object is ImmutableMappable
-			assert(false, "'ImmutableMappable' type requires throwing version of function \(#function)  - use 'try' before \(#function)")
+		} else if let klass = N.self as? ImmutableMappable.Type { // Check if object is ImmutableMappable
+			do {
+				return try klass.init(map: map) as? N
+			} catch let error {
+				#if DEBUG
+				let exception: NSException
+				if let mapError = error as? MapError {
+					exception = NSException(name: .init(rawValue: "MapError"), reason: mapError.description, userInfo: nil)
+				} else {
+					exception = NSException(name: .init(rawValue: "ImmutableMappableError"), reason: error.localizedDescription, userInfo: nil)
+				}
+				exception.raise()
+				#else
+				NSLog("\(error)")
+				#endif
+			}
 		} else {
 			// Ensure BaseMappable is not implemented directly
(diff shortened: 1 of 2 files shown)


tristanhimmelman:
Thank you @devxoul. I think this decision makes sense. 

Ideally developers just use one protocol throughout their app, but I understand that its not always that simple.

PR #781, opened 2017-03-21
fix indent depth

sohichiro:
This PR is not change original logic.
When I read your source, I realized that indent is not same.
Therefore I send PR for arrange it.


devxoul:
I saw @tristanhimmelman recently uses 4-spaces over tabs from AlamofireObjectMapper. Should we re-indent whole source code someday? 🤷‍♂️

tristanhimmelman:
@devxoul I don't have a strong preference with regards to indenting with spaces/tabs. Consistency is all that really matters to me. 

If you want to clean the indentation, feel free 👍

PR #785, opened 2017-03-29
Insert an ‘assert’, which may point out Generic Type parameter’s unconformity when debugging or testing.

Davidleeeeee:
Because I'm unfamiliar with the whole ObjectMapper, I only deal a part of the situation of operators' usage. I will do more later.


Davidleeeeee:
Actually, $0 is String Type.So when I define it as Float Type, the Assert work

PR #791, opened 2017-04-04
Abstract map key to MapKeyConvertible

mwoollard:
It'd be great if you would consider this PR which abstracts the mapping key to a new protocol `MapKeyConvertible`. `String` is extended to implement `MapKeyConvertible` so `String` keys can be used as before and thus breaking changes to existing users. `RawRepresentible` is also extended to implement `MapKeyConvertible` and so you can define your keys as an enum which is my main motivation behind the change:
```
enum Keys: String, MapKeyConvertible {
    case key1
    case key2
}
```
Having done this you can then use the enum as mapping key throughout `ObjectMapper`, a quick example:
```
    let value: String = try map.value(Keys.key1)
    let value: Int = try map.value(Keys.key2)
```
Let me know what you think :-)
Kind regards
Mark

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -35,16 +35,16 @@ script:
 
   # Build Framework in Debug and Run Tests if specified
   - if [ $RUN_TESTS == "YES" ]; then
-      xcodebuild -workspace "$WORKSPACE" -scheme "$SCHEME" -sdk "$SDK" -destination "$DESTINATION" -configuration Debug ONLY_ACTIVE_ARCH=NO test | xcpretty -c;
+      travis_retry xcodebuild -workspace "$WORKSPACE" -scheme "$SCHEME" -sdk "$SDK" -destination "$DESTINATION" -configuration Debug ONLY_ACTIVE_ARCH=NO test | xcpretty -c;
     else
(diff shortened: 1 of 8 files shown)


devxoul:
Can you provide some details? I don't think that is different from:

```swift
struct Keys {
  static let key1 = "key1"
  static let key2 = "key2"
}

let value: String = try map.value(Keys.key1)
let value: Int = try map.value(Keys.key2)
```

mwoollard:
Personally prefer `enum` approach to `struct` although agree it achieves the same. However by abstracting to protocol adds flexibility so that other implementations could be used if anyone wanted.
```
extension Int: MapKeyConvertible {
     public func asMapKey() -> String {
        return "value\(self)"
    }
}

//...

var array = [Int]()
(0..<10).forEach {
    array.append(try map.value($0))
}
```
It doesn't break existing code nor add any overhead in the performance unit tests when run locally (travis seems to vary run to run) and adds flexibility for anyone who wants it :-). I did encounter travis random failure with unit testing (simulator startup issue it seems) and so also just added a potential work around to the travis config.

Anyway let me know your thoughts and whether you want to include this feature...

Thanks
Mark

tristanhimmelman:
Thanks @mwoollard this is an interesting contribution. I am going to put some more thought into it as I make a big effort to keep the code base as simple as possible. Either way, I definitely want to incorporate your travis fix :)

PR #801, opened 2017-04-25
Prevent integer overflow crashes

bdbergeron:
The current implementation of integer mapping does not guard against integer overflow, and in addition, does not properly guard against trying to store 64-bit values in (U)Int on 32-bit devices. This is an attempt to mitigate against integer overflow crashes and unwrap values from JSON more appropriately into the specified (U)Int widths. This takes advantage of some new additions in Swift 3.1.


tristanhimmelman:
Thanks!

PR #802, opened 2017-04-26
fix HexColorTransform

KiuShuo:
Because:
UIColor.white.cgColor.numberOfComponents is 2.

### DIFF
--- a/Sources/HexColorTransform.swift
+++ b/Sources/HexColorTransform.swift
@@ -54,10 +54,15 @@ open class HexColorTransform: TransformType {
 	
 	fileprivate func hexString(color: Object) -> String {
 		let comps = color.cgColor.components!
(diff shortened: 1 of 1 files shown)


Sources/HexColorTransform.swift
```diff
@@ -54,10 +54,15 @@ open class HexColorTransform: TransformType {
 	
 	fileprivate func hexString(color: Object) -> String {
 		let comps = color.cgColor.components!
+		let compsCount = color.cgColor.numberOfComponents
```
Mazyod: My 2¢ here would be to `switch` on the number of components instead. It will better convey the purpose of the branching logic and how it is handled:

```
let r: Int
let g: Int
let b: Int
let a: Int
switch compsCount {
case 2: // Grayscale
    r = ..
    ...
case 4: // rgba
   r = ...
}
```

PR #810, opened 2017-05-13
Add a new class for cache

zzjzz9266a:
Cache an object or array to sandbox as json, easy to use.


devxoul:
I think it's not ObjectMapper's responsibility 🤔

tristanhimmelman:
Sorry but I agree with @devxoul. I don't think this needs to be incorporated into ObjectMapper. Thanks for the contribution

PR #817, opened 2017-05-26
Added Second DateFormatter and New Map Method

HassanEskandari:
In DateFormatterTransform, Added dateFormatter2 for cases when data received from remote server may vary in format

In Mapper class, added a new method to map JSON dictionary, accepting [String: Any?], which uses default map function by omitting null values from JSON dictionary


devxoul:
I think it's not a general purpose. It would be better to create a custom transform in your project.

tristanhimmelman:
I agree with @devxoul. Thanks for your contribution @HassanEskandari

PR #839, opened 2017-06-27
Map array of raw representables for ImmutableMappable

KIlleX00:
I have noticed that you can map `Raw Representable` type to a property without using `EnumTransform`, but when using array of raw representables you need to specify transform parameter otherwise mapping will fail.

Example:
```
let types: [TestType]    //TestType is Enum
let type: TestType

init(map: Map) throws {
    type = try map.value("type")
    types = try map.value("types")    // This will fail
} 
```

This is only valid for `ImmutableMappable`, on `Mappable` this already works without specifying tranform parameter. To be more consistent I have created this PR which solves this issue.


tristanhimmelman:
Thanks for the addition @KIlleX00 👍

cody1024d:
This is effecting my project right now as well.  Any ETA?  Is there anything I can do to help?

tristanhimmelman:
@cody1024d I just released a new version 2.2.8 with the above fix.

PR #846, opened 2017-07-04
- Mapper have to take into account Map shouldIncludeNilValues property

anton-plebanovich:
I had an issue when tried to make a JSON from an array of objects that contained another array of objects. Like that:
```Mapper(shouldIncludeNilValues: true).toJSONArray(objects)```

Inner array didn't include `nil` values.


tristanhimmelman:
Thanks for the fix!

PR #876, opened 2017-08-23
NSException raise if not Linux

conqueror:
This PR includes the following
- Fix for the Linux Build as NSException is not present in Linux Foundation
- Re-Indentation

ObjectMapper is tested with Perfect Framework and it works as expected.


conqueror:
It is not relevant anymore.

Issue #881, opened 2017-09-02
Map ignoreNil property


Arnoymous:
### Problem

I have a Mappable model with multiple subclasses, this is what I need to do:
create a method in my model that call the mapping(map:) method for any of its subclasses with a variation: full update (override) or just update with JSON's keys (useful for **Database**)

### Your JSON dictionary:

```json
{
  "name": "ObjectMapper",
  "url": "https://github.com/Hearst-DD/ObjectMapper"
}
```

### Your model:

```swift

struct Repo: Mappable {
  var name: String!
  var url: URL!

  mutating func update(withJSON JSON: [String: Any], fullUpdate: Bool) {
    let context: MapContext? = fullUpdate ? FullUpdateContext() : nil
    _ = Mapper(context: context).map(JSON: JSON, toObject: self)
  }

  mutating func mapping(map: Map) {
    name <- map["name", ignoreNil: !(map.context is FullUpdateContext)]
    url <- map["url", ignoreNil: !(map.context is FullUpdateContext)]
  }
}
```

### What you expected:

I exepected something like:

```swift

struct Repo: Mappable {
  var name: String!
  var url: URL!

  mutating func update(withJSON JSON: [String: Any], fullUpdate: Bool) {
    _ = Mapper(ignoreNil: !fullUpdate).map(JSON: JSON, toObject: self)
  }

  mutating func mapping(map: Map) {
    name <- map["name"]
    url <- map["url"]
  }
}
```

Arnoymous:
#882

PR #882, opened 2017-09-02
add ignoreNil property

Arnoymous:
#881


Arnoymous:
I removed test testIgnoreNilField()
The test pass because
```swift
Mapper<User>.parseJSONStringIntoDictionary(JSONString: JSON) = nil
```
and not 
```swift
firstName <- map["first_name", ignoreNil: false]
```

PR #888, opened 2017-09-18
Convert to Swift4.

RockerHX:
Convert substring(from:) to use String slicing subscript with a 'partial range from' operator.


tristanhimmelman:
Thanks for the fix but I incorporated this myself before seeing your PR.

PR #959, opened 2018-04-13
Issue with parsing Floats in Swift 4.1 

bilalbhutta:
Swift 4.1 breaks the Float parsing, ObjectMapper test cases were getting failed for floats. 
Issue is happening for the following data types 
1. Float
2. [Float] 
3. [String:Float]

### DIFF
--- a/Sources/Map.swift
+++ b/Sources/Map.swift
@@ -122,6 +122,7 @@ public final class Map {
 	public func value<T>() -> T? {
 		let value = currentValue as? T
 		
+		// Swift 4.1 breaks Float casting from `NSNumber`. So Added extra checks for `Flaot` `[Float]` and `[String:Float]`
 		if value == nil && T.self == Float.self {
 			if let v = currentValue as? NSNumber {
 				return v.floatValue as? T
--- a/Sources/Map.swift
+++ b/Sources/Map.swift
@@ -120,9 +120,23 @@ public final class Map {
 	}
 	
 	public func value<T>() -> T? {
-		return currentValue as? T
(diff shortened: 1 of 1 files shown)


pmusolino:
News?

samvanderhyden:
This is a critical issue for us. Any update would be appreciated!

bilalbhutta:
I'm using this branch for temporary fix.

> pod 'ObjectMapper', :git => 'https://github.com/bilalbhutta/ObjectMapper.git'

tristanhimmelman:
Very sorry for the delay on this. I have not had any time for maintenance on this project lately. Thanks for the PR

Sources/Map.swift
```diff
@@ -120,9 +120,24 @@ public final class Map {
 	}
 	
 	public func value<T>() -> T? {
-		return currentValue as? T
+		let value = currentValue as? T
+		
+		// Swift 4.1 breaks Float casting from `NSNumber`. So Added extra checks for `Flaot` `[Float]` and `[String:Float]`
```
Gerst20051: Typo has `Flaot` instead of `Float`

ashtiani:
This issue hasn't resolve yet, I updated my pod to latest version but I still see this issue. 
▿ 3 elements
  ▿ 0 : 2 elements
    - key : "ratingPercentage"
    - value : 89
  ▿ 1 : 2 elements
    - key : "rating"
    - value : 4.4444
  ▿ 2 : 2 elements
    - key : "count"
    - value : 9

this is what I get when I print the data, but after mapping the value of rating would be 0.

PR #976, opened 2018-06-05
emergency fix compile error for swift 4.2

manuqiao:


sunshinejr:
Thanks @manuqiao, wanted to do it myself and I saw that you already have done it 😉 Didn't have time to test it, but removing duplicated operators seems like a good fix. 

So what happened in Swift 4.2 is a change in Implicitely Unwrapped Optional (IUO). An overview on the changes in Swift 4.1 (where this issue was a warning) and in 4.2 (where this issue is now an error) is in [this blog post](https://swift.org/blog/iuo/).

And @JacobKingDev, please don't judge just by the code, sometimes there are different reason for a change and this seems like a good one (at least in my POV).

JacobKingDev:
@sunshinejr My point was more a case of simply removing code will cause issues for users of the library that utilised such code, not to mention potentially causing compiler errors where such methods are used within the library. Also, I noticed builds are failing but on closer inspection I think this is because CI is configured to use an outdated Swift version. If this fix is good, can someone merge it? I'm champing at the bit to use dark mode! 😛

sunshinejr:
@JacobKingDev sure, it just seemed a bit rude, if you'd write it the way you explained it now, I wouldn't say anything. But no worries, it's all good 👍

About the removals, I think it seems okayish, because these are just redundant (after the changes the method will accept optional either way, no need for unforced unwrapped optional as an argument). I'm not sure for 100%, but it seems reasonable.

talesp:
Maybe a better/safer solution would be something like
```swift
#if swift(>=4.1)
#else
/// Implicitly Unwrapped Optional Object of Raw Representable type
public func <- <T: RawRepresentable>(left: inout T!, right: Map) {
	left <- (right, EnumTransform())
}
#endif
```
(Sadly swift version check doesn't work fot less than checks)

sunshinejr:
@talesp Unfortunately this won't work as this is also changed in 3.* version that Xcode 10 ships with. Thus, it won't compile as well. 

We would need the [`compiler` directive](https://github.com/apple/swift-evolution/blob/master/proposals/0212-compiler-version-directive.md), but that was just merged yesterday so we would need to wait for next beta probably.

sunshinejr:
New to me as well. I've heard the news about 4.2 but not 3.*. But it kinda _makes sense_ as they wanted to remove the IUO implementation from the codebase, so. 

Tried it yesterday and today just to be sure and it won't compile when you use Swift 3 for the target - still throws an error.

talesp:
yeah, but breaks compatibility to an already released version :(
So, maybe a `xcode-10` branch for now, breaking compatibility with Xcode 9 (that does not have the `compiler` directive)?

tristanhimmelman:
Thanks for the contribution but I already merged another swift 4.2 PR.

PR #980, opened 2018-06-07
Added support for Swift 4.2 and Xcode 10

gcharita:
Contents of this pull request:
- Added support for Swift 4.2 without dropping support for Xcode 9.3 (and Xcode 9.4) and Swift 4.1 (in Xcode 9.3 and 9.4 warnings will remain due to the fact that the behaviour of IUO is not the same as Xcode 10 beta and as a result compilation fails in tests)
- Also fixed support for Xcode 9.2 and swift 4.0.

PR #993, opened 2018-08-23
Hotfix/swift 4 2 compiler warnings

f2m2rd:


gcharita:
As mentioned in #980, we are getting compilation error `Binary operator '<-' cannot be applied to operands of type` when building Test targets with Xcode 9.4.1

screen shot 2018-08-24 at 11 35 41

Issue #402, opened 2016-03-02
DateTransform not Converting milliseconds to Date accurately


umairalighumman:
I am using objectmapper to parse json to model. when i parse millisecond to date it could not accurately calculate date. I resolve this issue by dividing 1000 to millisecond in Date Transform  .

tristanhimmelman:
`DateTransform` currently requires an input of seconds not milliseconds. 

If you want to add functionality to DateTransform so that it can optionally support milliseconds, I would be happy to accept a Pull Request. You could add an optional parameter to the initializer that configures the transform to operate on milliseconds.

PR #1015, opened 2018-10-07
DateTransformer: Support milliseconds

Mazyod:
Add milliseconds unit support for `DateTransformer`, as suggested in #402


Mazyod:
@tristanhimmelman It has tests, and I've also integrated it with our app, and seems to be working fine.

Please LMK if there are any changes on your mind, and I can accommodate.

tristanhimmelman:
Thanks @Mazyod

PR #1036, opened 2018-12-17
Add failing test for nested dictionary with custom mappables

edenman:
This seems like a bug to me, but maybe I'm just doing something wrong?  If you can point me either to what I'm doing wrong or where the bug is in the source, I'm happy to put together a PR with either a fix for the bug (or if it's just me doing something wrong, a fix for this PR as an example test for future custom-Mappable-builders).

### DIFF
--- a/Tests/ObjectMapperTests/MappableExtensionsTests.swift
+++ b/Tests/ObjectMapperTests/MappableExtensionsTests.swift
@@ -57,6 +57,60 @@ func ==(lhs: TestMappable, rhs: TestMappable) -> Bool {
 	return lhs.value == rhs.value
 }
 
+class CustomDictionary<KeyType: Hashable, ValueType>: Mappable {
+	var _dictionary: [KeyType: ValueType]
+
+	init() {
+		_dictionary = [:]
+	}
+
+	required init?(map: Map) {
+		_dictionary = [:]
+	}
+
+	func mapping(map: Map) {
+		_dictionary <- map["dict"]
+	}
+
+	subscript(key: KeyType) -> ValueType? {
+		get {
+			return _dictionary[key]
+		}
+		set {
+			if newValue == nil {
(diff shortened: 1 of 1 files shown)


edenman:
@tristanhimmelman why'd this get closed? i'd love to actually get this bug fixed

Tests/ObjectMapperTests/MappableExtensionsTests.swift
```diff
@@ -57,6 +57,60 @@ func ==(lhs: TestMappable, rhs: TestMappable) -> Bool {
 	return lhs.value == rhs.value
 }
 
+class CustomDictionary<KeyType: Hashable, ValueType>: Mappable {
```
tristanhimmelman: Below you are passing in a `Mappable` type for `ValueType`. If you expect `ValueType` to map, it needs to conform to `Mappable` in the class definition. 

Not 100% sure, but this might fix your issue:
`class CustomDictionary<KeyType: Hashable, ValueType: Mappable>: Mappable`

edenman: No luck :(

edenman: The debugger doesn't like all the custom operators but I added a bunch of NSLogs and finally figured out the issue:
```
public func <- <T: BaseMappable>(left: inout Dictionary<String, T>, right: Map) {
```
My custom map doesn't specify that the keys have to be `String`s, just `Hashable`s.  Which means the Dictionary operator doesn't get selected, and instead it uses the "Object of Basic type" one and that one doesn't transform the `TestMappable`s to json, which makes `isValidJSONObject` return false and `toJSONString` ends up returning nil.

Is this just a "make CustomDictionary have Strings for keys" situation?  I can't really think of a situation where non-String keys would make sense.

PR #1057, opened 2019-06-03
Fix #1043 thanks to @cddjr 

### LINKED ISSUE #1043: Swift 5 and try?
Parsing complicated object with optional objects as a parameter does not work.

```
public struct Push: ImmutableMappable {
  public let contact: MeetingContactPush?

  public init(map: Map) throws {
    contact = try? map.value("Contact")
  }

}

public struct MeetingContactPush: ImmutableMappable {

  public let userId: Int
  public let phone: String
  public let name: String?

  public init(map: Map) throws {
    userId = try map.value("Id")
    phone = try map.value("Phone")
    name = try? map.value("Name")
  }

}

```
Contact will be always nil

mickamy:
Original fixing-code was written by @cddjr at https://github.com/cddjr/ObjectMapper/commit/2066cb86d89aba890f7067a3f8e546acf1474c9e

I just copy and pasted his code and add some comment.
All the honor for this commit should be given to @cddjr.

I just want this library to be updated with that fix.
Though he/she did not open PR for months, so I made it.

### DIFF
--- a/Sources/ImmutableMappable.swift
+++ b/Sources/ImmutableMappable.swift
@@ -99,11 +99,32 @@ public extension Map {
 		}
 		return try Mapper<T>(context: context).mapOrFail(JSONObject: JSONObject)
 	}
+	
+	/// Returns a `BaseMappable` object boxed in `Optional` or throws an error.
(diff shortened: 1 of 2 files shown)


pmusolino:
Thanks, @mickamy! I hope that @tristanhimmelman will be merge this PR soon.

PR #1061, opened 2019-06-14
move Mapper static func to MapperParse

yixiaojichunqiu:
later code,when i use `public static func toJSONString(_ JSONObject: Any, prettyPrint: Bool) -> String?` very inconvenient. Mapper use generic and these func will happen Generic parameter 'N' could not be inferred.I think these static func not in Mapper will be better.

### DIFF
--- a/Sources/MapperParse.swift
+++ b/Sources/MapperParse.swift
@@ -1,0 +1,64 @@
+//
+//  MapperParse.swift
+//  ObjectMapper-iOS
+//
+//  Created by yixiaojichunqiu on 2019/6/13.
+//  Copyright © 2019 hearst. All rights reserved.
+//
+
+import Foundation
+
+public final class MapperParse
+{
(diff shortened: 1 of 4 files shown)


Sources/MapperParse.swift
```diff
@@ -0,0 +1,64 @@
...
+
+import Foundation
+
+public final class MapperParse
+{
+	/// Converts an Object to a JSON string with option of pretty formatting
+	public static func toJSONString(_ JSONObject: Any, prettyPrint: Bool) -> String? {
+		let options: JSONSerialization.WritingOptions = prettyPrint ? .prettyPrinted : []
+		if let JSON = MapperParse.toJSONData(JSONObject, options: options) {
+			return String(data: JSON, encoding: String.Encoding.utf8)
```
emrcftci: Please pass just `.utf8` instead of `String.Encoding.utf8`

 `return String(data: JSON, encoding: .utf8)`

PR #1066, opened 2019-07-11
Ubdate to Swift 5.0

savoptik:
I   updated the code   to   Swift   5.0

### DIFF
--- a/Sources/ImmutableMappable.swift
+++ b/Sources/ImmutableMappable.swift
@@ -272,7 +272,7 @@ public extension Mapper where N: ImmutableMappable {
 		#if swift(>=4.1)
 		return try JSONArray.compactMap(mapOrFail)
 		#else
-		return try JSONArray.flatMap(mapOrFail)
+		return try JSONArray.compactMap(mapOrFail)
 		#endif
 	}
 	
--- a/Sources/Map.swift
+++ b/Sources/Map.swift
@@ -134,7 +134,7 @@ public final class Map {
 				#if swift(>=4.1)
 				return v.compactMap{ Float($0) } as? T
 				#else
-				return v.flatMap{ Float($0) } as? T
+				return v.compactMap{ Float($0) } as? T
 				#endif
 			}
 		} else if value == nil && T.self == [String:Float].self {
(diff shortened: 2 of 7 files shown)


Sources/ImmutableMappable.swift
```diff
@@ -272,7 +272,7 @@ public extension Mapper where N: ImmutableMappable {
 		#if swift(>=4.1)
 		return try JSONArray.compactMap(mapOrFail)
 		#else
-		return try JSONArray.flatMap(mapOrFail)
+		return try JSONArray.compactMap(mapOrFail)
```
emrcftci: Please just remove `else` statement if it is necessary because now `if` and `else` blocks has same actions.

Sources/Map.swift
```diff
@@ -134,7 +134,7 @@ public final class Map {
 				#if swift(>=4.1)
 				return v.compactMap{ Float($0) } as? T
 				#else
-				return v.flatMap{ Float($0) } as? T
+				return v.compactMap{ Float($0) } as? T
```
emrcftci: **Same comment as above.**

Please just remove `else` statement if it is necessary because now `if` and `else` blocks has same actions.

Sources/Mapper.swift
```diff
@@ -165,7 +165,7 @@ public final class Mapper<N: BaseMappable> {
 		#if swift(>=4.1)
 		let result = JSONArray.compactMap(map)
 		#else
-		let result = JSONArray.flatMap(map)
+		let result = JSONArray.compactMap(map)
```
emrcftci: **Same comment as above.**
Please just remove `else` statement if it is necessary because now `if` and `else` blocks has same actions.

Sources/Mapper.swift
```diff
@@ -433,7 +433,7 @@ extension Mapper where N: Hashable {
 		#if swift(>=4.1)
 		return Set(JSONArray.compactMap(map))
 		#else
-		return Set(JSONArray.flatMap(map))
+		return Set(JSONArray.compactMap(map))
```
emrcftci: **Same comment as above.**
Please just remove `else` statement if it is necessary because now `if` and `else` blocks has same actions.

PR #1079, opened 2019-12-03
ImmutableMappable init fix

itruf:
I've started research to fix warning mentioned here:
https://github.com/tristanhimmelman/ObjectMapper/issues/1043#issuecomment-479406127

If you try to init ImmutableMappable with JSONString - swift shows warning that this method has no calls to throwing functions.

I found that:
- `ImmutableMappable` inherits from `BaseMappable`
- `ImmutableMappable` has an extension with init methods which throws (like `init(JSONString: String, context: MapContext? = nil) throws`)
- `BaseMappable` has an extension with optional init methods (like `init?(JSONString: String, context: MapContext? = nil)`)

It causes crashes in some cases.

For example:
`let response = try? MyModel(JSONString: jsonString)` will crash if `MyModel` is ImmutableMappable and throws in init method.

I'm not sure but it can be caused by selecting optional init from `BaseMappable` which not marked with `throws`.

### DIFF
--- a/Sources/Mappable.swift
+++ b/Sources/Mappable.swift
@@ -46,7 +46,7 @@ public protocol StaticMappable: BaseMappable {
 	static func objectForMapping(map: Map) -> BaseMappable?
 }
 
-public extension BaseMappable {
+public extension Mappable {
 	
 	/// Initializes object from a JSON String
 	init?(JSONString: String, context: MapContext? = nil) {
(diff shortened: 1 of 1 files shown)


itruf:
@tristanhimmelman I'm not pretty sure about library architecture. Hope that you will find some time to check my PR.

itruf:
@tristanhimmelman it's a critical problem with ImmutableMappable. Please check it out.

itruf:
@tristanhimmelman I'm really sorry for pushing you but it's the very important PR for the project. I hope that you will take a look at it.

itruf:
Thank you @tristanhimmelman

murad1981:
this update caused many crashes in my app where I init an ImmutableMappable from a json string and from a json dict, both now cause runtime crash.

itruf:
@murad1981 can you show an example of your model and JSON string? I think that it's correct behavior if you init an ImmutableMappable model with an incorrect JSON.

PR #1080, opened 2019-12-12
Be able to use own encoder and decoder.

noppefoxwolf:


Sources/CodableTransform.swift
```diff
@@ -33,8 +33,13 @@ open class CodableTransform<T: Codable>: TransformType {
 
     public typealias Object = T
     public typealias JSON = Any
-
-    public init() {}
+    let encoder: JSONEncoder
+    let decoder: JSONDecoder
+    
+    public init(decoder: JSONDecoder = .init(), encoder: JSONEncoder = .init()) {
```
emrcftci: Why you decided to create an initializer for these properties?

I think this development is unnecessary to use `encoder` and `decoder` because these properties aren't use anywhere except functions.

On the other hand some cases we don't need to create `encoder` or `decoder`.
For example; if `data` comes NULL in `transformFromJSON (_ value:) -> Object?` function returns immediately from `guard` statement and we don't create a `decoder`.

`transformToJSON (_ value: ) -> JSON?` has a same case, if `value` comes NULL we don't create `encoder`.

What do you think about this cases?

noppefoxwolf: We want to set options for JSONDecoder and JSONEncoder.
For example, cannot set dataDecodingStrategy by current code.

Sources/CodableTransform.swift
```diff
@@ -33,8 +33,13 @@ open class CodableTransform<T: Codable>: TransformType {
 
     public typealias Object = T
     public typealias JSON = Any
-
-    public init() {}
+    let encoder: JSONEncoder
+    let decoder: JSONDecoder
```
gcharita: Maybe you should consider making those properties private. What do you think?

gcharita (changes requested):

Sources/CodableTransform.swift
noppefoxwolf: YES, serializers should be private.

Sources/CodableTransform.swift
```diff
@@ -33,8 +33,13 @@ open class CodableTransform<T: Codable>: TransformType {
 
     public typealias Object = T
     public typealias JSON = Any
-
-    public init() {}
+    let encoder: JSONEncoder
```
noppefoxwolf:
```suggestion
    private let encoder: JSONEncoder
```

Sources/CodableTransform.swift
noppefoxwolf:
```suggestion
    private let decoder: JSONDecoder
```

gcharita (approved):

PR #1086, opened 2020-03-11
Removing recursive linking to get Object Mapper working w/ Xcode 11.4's Swift Package Manager

ksloginov:
**Description**
Simplifying the folder structure to avoid double-linking of the same classes.

**Motivation and Context**
With Xcode 11.4 the new set of rules have been introduced to SPM and now the older version of ObjectMapper can't compile with the latest Xcode (if added via SPM)

**Issue-Ticket**:
https://github.com/tristanhimmelman/ObjectMapper/issues/1082

[+] Bug fix (non-breaking change which fixes an issue)
[+] I have run tests


.swiftpm/xcode/xcshareddata/xcschemes/ObjectMapper.xcscheme
```diff
@@ -0,0 +1,67 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
SergeyPetrachkov: should it be in the repo at all?

joshuapoq: I don't think it needs to be committed; I'm seeing mixed practise with it but see no reason to maintain these schemes as they get regenerated 🤔

PR #1116, opened 2020-12-29
Add Support for Decimal Type

hanleylee:
## Why 

For avoiding the precision loss in Swift, we want to use Decimal Type to accept data from Server side, but currently ObjectMapper don't support Decimal transform

## What 

add DecimalTransform and related Tests to support Decimal Type


alinfarcas12:
+1

PR #1131, opened 2023-12-14
Add Multi-Keys to single property support

uxinting:
Some times server api return different key but same meaning value results, in this case, we have to write if-else control to choose the valid one. I commit this to solve the situation.

If can't get value from primary choice, use this instead
first json data like:
```json
{
  username: "John Doe"
}
```
second json data like:
 ```json
{
  nickname: "John"
}
```
Usage:
```swift
func mapping(map: Map) {
	//...
 	name	<- map["username"].or["nickname"]
 	//...
}
```
You can get valid name either the first or the second json data.

## Change
add function `or` at the extension of Map, if map's `currentValue` is nil, use the `or's one`


ShenYj (commented):
good！

PR #1142, opened 2024-04-04
Update Apple Upcoming Requirements

selvam4274:
Privacy Manifests:
https://developer.apple.com/support/third-party-SDK-requirements/

App Store Connect upload requirement
https://developer.apple.com/news/?id=fxu2qp7b

## Why 

Screenshot 2024-04-04 at 9 18 47 AM

## What 

- Updated min target version 
- Added third-party-SDK-requirements

Test Case
Screenshot 2024-04-04 at 11 37 50 AM

Screenshot 2024-04-04 at 11 38 00 AM

Screenshot 2024-04-04 at 11 38 09 AM


elmetal:
Great! It seems that the empty privacy manifest might be unnecessary. What do you think?

> IS PrivacyManifest.xcprivacy still required for Framework even if no privacy accessed API is used? - https://forums.developer.apple.com/forums/thread/746481?answerId=782256022#782256022
>
> If your framework doesn't require a privacy manifest, do nothing. Avoid adding an empty privacy manifest to your framework.

selvam4274:
> Great! It seems that the empty privacy manifest might be unnecessary. What do you think?
> 
> > IS PrivacyManifest.xcprivacy still required for Framework even if no privacy accessed API is used? - https://forums.developer.apple.com/forums/thread/746481?answerId=782256022#782256022
> > If your framework doesn't require a privacy manifest, do nothing. Avoid adding an empty privacy manifest to your framework.

@elmetal Thanks for bringing this up. It's an interesting point regarding the necessity of the PrivacyManifest.xcprivacy file, especially, I referred Snapkit  

listed under https://developer.apple.com/support/third-party-SDK-requirements

Screenshot 2024-04-04 at 11 19 31 PM

With empty privacy manifest

Screenshot 2024-04-04 at 11 19 31 PM

We can avoid unnecessary app rejection or the need to appeal a review by simply not adding this.

ObjectMapper.podspec
```diff
@@ -1,16 +1,21 @@
...
-  s.ios.deployment_target = '13.0'
-  s.osx.deployment_target = '10.10'
-  s.tvos.deployment_target = '9.0'
+  s.resource_bundle = {
+    "Privacy" => "Sources/ObjectMapper/Resources/PrivacyInfo.xcprivacy"
+  }
+  s.watchos.deployment_target = '10.0'
+  s.ios.deployment_target = '17.0'
+  s.osx.deployment_target = '12.0'
+  s.tvos.deployment_target = '17.0'
```
anton-plebanovich, line 18: I assume "must be built with [Xcode 15](https://apps.apple.com/us/app/xcode/id497799835?mt=12) for [iOS 17](https://developer.apple.com/ios/)..." does not mean we should set the deployment target to iOS 17 as it will break most existing clients of the library who still might support earlier iOS versions.

If you check https://developer.apple.com/ios/submit/ it just says "must be built with a minimum of Xcode 15 and the iOS 17 SDK." There is no limitation to the deployment target it just has to be built with Xcode 15 and the latest SDK which makes sense.

I believe we will have to revert most of the introduced limitations in this pull request.

mvn-thanhnguyen4-dn: @tristanhimmelman 
This is correct. Can you revert this code so the library can support old iOS versions? Thank you

tristanhimmelman: @mvn-thanhnguyen4-dn that makes sense. Can you submit a PR please

mvn-thanhnguyen4-dn: @tristanhimmelman 
Please review this PR. Thank you
https://github.com/tristanhimmelman/ObjectMapper/pull/1146

Issue #1147, opened 2024-04-24
Current Package.swift config does not work


kaulex99:
The current `Package.swift` config is not working. When Xcode tries to resolve the ObjectMapper version 4.4.2 it ends up in following error. 
```
/Package.swift:7:43: error: 'v12' is unavailable
                      platforms: [.macOS(.v12),
                                          ^~~
PackageDescription.SupportedPlatform:21:27: note: 'v12' was introduced in PackageDescription 5.5
        public static let v12: PackageDescription.SupportedPlatform.MacOSVersion
                          ^ in https://github.com/tristanhimmelman/ObjectMapper.git
```

So the first line of the package file need to be changed to version 5.5
```
// swift-tools-version:5.5
```

@tristanhimmelman

kaulex99:
PR has been created and needs to be reviewed & merged.

PR #1148, opened 2024-04-25
Update Package.swift to swift-tools-version 5.5

kaulex99:
## Why 
Issue: #1147

The current Package.swift is not valid as `.macOS(.v12)` was introduced in 5.5.

## What
Changed the swift-tools-version to 5.5


alexanderkranga:
can someone please take a look and merge this? This is breaking our library builds

tristanhimmelman (approved):

